PR #48, opened 2023-02-03
feat: Added Support for Form Fields

holtskinner:
Changed `_text_from_element_with_layout()` to `_text_from_layout()` and removed Union to allow for FormField name/value parsing


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -331,6 +331,26 @@ def search_pages(
                     break
         return found_pages
 
+    def get_form_field_by_name(self, target_field: str):
```
galz10, line 334: add return type

galz10 (approved):
Other then the forgotten return type this looks good.

PR #57, opened 2023-02-14
test: Resolved Linter and Test Errors

holtskinner:
Resolves Type import error created in #48

### DIFF
--- a/google/cloud/documentai_toolbox/wrappers/document.py
+++ b/google/cloud/documentai_toolbox/wrappers/document.py
@@ -27,6 +27,7 @@ from google.cloud import documentai_toolbox
 
 from google.cloud.documentai_toolbox import constants
 from google.cloud.documentai_toolbox.wrappers.page import Page
+from google.cloud.documentai_toolbox.wrappers.page import FormField
 from google.cloud.documentai_toolbox.wrappers.entity import Entity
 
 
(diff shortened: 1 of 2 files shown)


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -332,15 +331,15 @@ def search_pages(
                     break
         return found_pages
 
-    def get_form_field_by_name(self, target_field: str) -> List[FormField]:
```
galz10, line 335: Please revert this change and add FormField to main __init__.py file

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -291,19 +291,19 @@ def _get_form_fields(
 
 
 def _table_rows_from_documentai_table_rows(
-    table_rows: List[documentai.Document.Page.Table.TableRow], text: str
-) -> List[str]:
+    table_rows: MutableSequence[documentai.Document.Page.Table.TableRow], text: str
```
galz10, line 294: Why do you need a MutableSequence vs a List?

holtskinner: mypy was saying that the types are incompatible. (Might be an implicit casting issue)

holtskinner: MutableSequence is the type that `documentai.Document.Page.Table.table_rows` is from the Document AI client library

PR #66, opened 2023-02-27
feat: Added Batch creation for Cloud Storage documents.

holtskinner:


holtskinner:
Also, general question on the utilities module. Is there a way to make it so that the public library can import utilities like this:

```py
from google.cloud.documentai_toolbox import utilities
```

instead of this:

```py
from google.cloud.documentai_toolbox.utilities import utilities
```

The Document file doesn't require `documentai_toolbox.wrappers` to import

samples/snippets/create_batches_sample.py
```diff
@@ -0,0 +1,49 @@
...
+from google.cloud.documentai_toolbox.utilities import utilities
+
+# TODO(developer): Uncomment these variables before running the sample.
+# Given unprocessed documents in path gs://bucket/path/to/folder
+# gcs_bucket_name = "bucket"
+# gcs_prefix = "path/to/folder"
+# batch_size = 50
+
+
+def create_batches_sample(
```
galz10, line 29: Nit: could you add another function that uses the batches to call documentai, i think it would be helpful to show how to use the batches.

galz10 (approved):

samples/snippets/create_batches_sample.py
holtskinner: I considered this, but it would be a bunch of duplicated code from the batch_process sample. This will be embedded in the docs on the same page that shows how to send a batch process request. I'll add a link

PR #69, opened 2023-03-01
fix: Handle Edge Case where GCS Shards are out of order

### LINKED ISSUE #68: Shard consolidation edge case - GCS Blobs may be in the different order than shardIndex
From @ffeldhaus

The edge cases I think may not be covered are, that blobs may be in a different order than their `shardIndex`.
I had to correct the page number by a page offset

#### Code example

```python
def merge_sharded_results(blobs: List[storage.Blob], input_bucket_name, input_object_name, result_bucket_name, result_prefix):
    shard_count = len(blobs)
    print(f"Merging {len(blobs)} sharded results")
    merged_document = {}
    page_image_names_per_shard = [None] * shard_count
    text_offset = 0
    shard_index = next_shard_index = 0
    text = [''] * shard_count
    for blob in blobs:
        print(f"Downloading content of blob {[blob.name](http://blob.name/)}")
        content = blob.download_as_bytes()
        document = json.loads(content)
        del content
        print(f"Extracting images to external files or render from input document")
        # page_offset is the minimum of the pageNumbers of the current shard but starts at 0 - thus substracting 1
        page_offset = min(int(page['pageNumber'])
                          for page in document['pages']) - 1
        if 'shardInfo' in document:
            shard_info = document['shardInfo']
            if 'textOffset' in shard_info:
                text_offset = int(shard_info['textOffset'])
            if 'shardCount' in shard_info:
                if shard_count != int(shard_info['shardCount']):
                    raise ValueError(f"Shard count mismatch! Current shard_count is {shard_count} shardInfo has {shard_info['shardCount']} and there are {len(blobs)} blobs with shard information")
            if 'shardIndex' in shard_info:
                shard_index = int(shard_info['shardIndex'])
                if shard_index >= shard_count:
                    raise ValueError(f"Shard index {shard_index} larger than or equal to shard count {shard_count}")
            else:
              # auto increment shard_index by one - only works if blobs are in the same order as shard index, which they usually are but not guaranteed
                next_shard_index = shard_index + 1
            del document['shardInfo']
        print(f"Working on shard_index {shard_index} out of {shard_count} total shards")
        print("Extracting images from current shard")
        page_image_names_per_shard[shard_index] = extract_image(document, input_bucket_name, input_object_name, result_bucket_name, result_prefix, page_offset)
        print("Remove images from document JSON")
        for page in document.get('pages',[]):
              if 'image' in page:
                del page['image']
        print("Applying page offset to all pageRefs")
        for entity in document.get('entities',[]):
            page_anchor = entity.get('pageAnchor')
            if page_anchor:
                page_refs = page_anchor.get('pageRefs')
                if page_refs:
                    for page_ref in page_refs:
                        page = page_ref.get('page')
                        if page:
                            page_ref['page'] = int(page) + page_offset
                        else:
                            page_ref['page'] = page_offset
        print("Applying text offset to all text segments")
        apply_text_offset(document, text_offset)
        print("Merging document into merged_document")
        for key in document.keys():
            if isinstance(document[key], list):
                if key in merged_document:
                    merged_document[key] = merged_document[key] + document[key]
                else:
                    merged_document[key] = document[key]
            elif key == "text":
                text[shard_index] = document[key]
            else:
                if key not in merged_document:
                    merged_document[key] = document[key]
        del document
        gc.collect()
        print("Memory: ", psutil.virtual_memory())
        shard_index = next_shard_index
    # flatten page_image_names_per_shard
    page_image_names = [item for sub_list in page_image_names_per_shard if sub_list for item in sub_list]
    # join all text entries in the right order
    merged_document['text'] = ''.join(text)
    return merged_document, page_image_names
```

I had to apply text offsets to all textSegments
and I had to do garbage collection after each shard otherwise I was running out of memory on documents with 100s of pages

```python
def apply_text_offset(json_input, text_offset):
    lookup_key = "textSegments"
    if isinstance(json_input, dict):
        for k, v in json_input.items():
            if k == lookup_key:
                for text_segment in v:
                    if "startIndex" in text_segment:
                        text_segment['startIndex'] = str(
                            int(text_segment['startIndex']) + text_offset)
                    if "endIndex" in text_segment:
                        text_segment['endIndex'] = str(
                            int(text_segment['endIndex']) + text_offset)
            else:
                apply_text_offset(v, text_offset)
    elif isinstance(json_input, list):
        for item in json_input:
            apply_text_offset(item, text_offset)
```
Feel free to use any part of the code I shared. I'd be glad if we have an official solution to merge shards into a single JSON

holtskinner:
Fixes #68 🦕


tests/unit/resources/unordered_shards/Winnie_the_Pooh-0.json
```diff
@@ -0,0 +1 @@
...into the Cunning Trap to\ncatch the Heffalump.\n\"Bother!\" said Pooh. \"It all comes of trying to be\nkind to Heffalumps.\" And he got back into bed.\nBut he couldn't sleep. The more he tried to sleep,\nthe more he couldn't. He tried Counting Sheep,\nwhich is sometimes a good way of getting to sleep,\nand, as that was no good, he tried counting Heffa-\nlumps. And that was worse. Because every Heffa-\nlump that he counted was making straight for a pot\nDigitized by\nGoogle\nPIGLET MEETS A HEFFALUMP\nof Pooh's honey, and eating it all. For some min-\nutes he lay there miserably, but when the five hun-\ndred and eighty-seventh Heffalump was licking its\njaws, and saying to itself, \"Very good honey this,\nI don't know when I've tasted better,\" Pooh could\nbear it no longer. He jumped out of bed, he ran out\nof the house, and he ran straight to the Six Pine\nTrees.\n63\nThe Sun was still in bed, but there was a light-\nness in the sky over the Hundred Acre Wood\nwhich seemed to show that it was waking up and\nwould soon be kicking off the clothes. In the half-\nLight the Pine Trees looked cold and lonely, and the\nVery Deep Pit seemed deeper than it was, and\nDigitized by\nGoogle\n"}
```
galz10, line 1: Nit : could you reduce the number of unordered shards

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -160,6 +160,7 @@ def _get_shards(gcs_bucket_name: str, gcs_prefix: str) -> List[documentai.Docume
     for byte in byte_array:
         shards.append(documentai.Document.from_json(byte, ignore_unknown_fields=True))
 
+    shards.sort(key=lambda x: int(x.shard_info.shard_index))
```
galz10, line 163: Does every document.proto had a shard index even if it's not sharded ? What will happen if the document that's passed in isn't sharded?

holtskinner: Unsharded documents won't have a shard index, but it will have `shard_info`.

`int(x.shard_info.shard_index)` will evaluate to `0`. Plus if there's only one shard, sorting won't do anything. Would you prefer I add a condition to check if it is multisharded first?

galz10: Yea i think validating that it's multiple would be better.

galz10 (approved):

PR #75, opened 2023-03-09
feat: added list_gcs_document_tree

galz10:

### DIFF
--- a/google/cloud/documentai_toolbox/utilities/utilities.py
+++ b/google/cloud/documentai_toolbox/utilities/utilities.py
@@ -24,8 +24,8 @@ from google.cloud.documentai_toolbox import constants
 from google.cloud.documentai_toolbox.wrappers.document import _get_storage_client
 
 
-def print_gcs_document_tree(gcs_bucket_name: str, gcs_prefix: str) -> None:
-    r"""Prints a tree of filenames in Cloud Storage folder.
+def list_gcs_document_tree(gcs_bucket_name: str, gcs_prefix: str) -> List[str]:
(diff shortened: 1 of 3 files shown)


google/cloud/documentai_toolbox/utilities/utilities.py
```diff
@@ -24,8 +24,8 @@
 from google.cloud.documentai_toolbox.wrappers.document import _get_storage_client
 
 
-def print_gcs_document_tree(gcs_bucket_name: str, gcs_prefix: str) -> None:
-    r"""Prints a tree of filenames in Cloud Storage folder.
```
holtskinner, line 28: I think it makes sense to return a list of documents, but I also think that there should be a separate function for printing (and not have the list function print it). Just seems like each function should only have one purpose

Maybe the print function could call `list_gcs_document_tree` in the background, then just print the list.

google/cloud/documentai_toolbox/utilities/utilities.py
```diff
@@ -73,6 +99,8 @@ def print_gcs_document_tree(gcs_bucket_name: str, gcs_prefix: str) -> None:
             elif idx <= FILES_TO_DISPLAY:
                 print(f"{FILENAME_TREE_MIDDLE}{file_name}")
 
+    return list(path_list.keys())
```
holtskinner, line 102: This function shouldn't return anything

holtskinner (approved):

google/cloud/documentai_toolbox/utilities/utilities.py
galz10: ah good catch

holtskinner (approved):

PR #110, opened 2023-04-19
refactor: Restructuring of classes to reduce duplication and increase readibility

holtskinner:
- Changed `vision_helpers.py` to input `Document.Page` to reduce extra imports.
- Rewrote Entity image extraction to avoid storing the Pillow image in memory.
- Added error checking for missing shards.
- Added sorting of shards by index in case they are read out of order.
- Fixed bug where `Entity` page numbers would be incorrect in multi-shard documents.
- Simplified creation of `Page` sub classes by using `__post_init__()`
- Renamed `Page.text` and `Table.text` to `document_text` to be more descriptive and match the other `Page` elements.
- Rewrote `table_sample.py` to use more accurate/descriptive variable names.
- Minor formatting of comments/docs


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -717,15 +695,34 @@ def split_pdf(self, pdf_path: str, output_path: str) -> List[str]:
...
             AnnotateFileResponse:
-                Proto with TextAnnotations.
+                Proto with `TextAnnotations`.
         """
-        return _convert_to_vision_annotate_file_response(self.text, self.pages)
+        responses: List[AnnotateImageResponse] = []
+
+        for shard in self.shards:
+            for docai_page in shard.pages:
+                page_vision_annotation = _convert_document_page(
```
dizcology, line 710: Can we take this opportunity to import module instead of functions?  In particular this would become `vision_helpers._convert_document_page`, making its use cases more obvious.

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -64,17 +62,19 @@ def _entities_from_shards(
...
-            entity.crop_image(shard)
-            result.append(entity)
-            for documentai_prop in documentai_entity.properties:
-                prop = Entity(documentai_entity=documentai_prop)
-                prop.crop_image(shard)
-                result.append(prop)
-
-    if len(result) > 1 and result[0].documentai_entity.id:
-        result.sort(key=lambda x: int(x.documentai_entity.id))
+        shard_index = int(shard.shard_info.shard_index)
```
dizcology, line 65: Do we know that `shard.shard_info.shard_index` will always be the same as the index of `shard` as it appears in the list `shards`?

In `export_images` below we have the following code:
```
image = entity.crop_image(
                documentai_document=self.shards[entity.shard_index]
            )
```
which suggests that these two notion of "shard index" are indeed the same.  Can we confirm that to be the case?

Alternatively, from the perspective of the Toolbox library, the "index of a shard as it appears in the list shards" is the one actually used, so perhaps we can simply iterate with `for shard_index, shard in enumerate(shards): ...` here instead.

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -26,44 +27,73 @@
 
 @dataclasses.dataclass
 class Entity:
-    r"""Represents a wrapped documentai.Document.Entity.
+    """Represents a wrapped `documentai.Document.Entity`.
 
     Attributes:
+        shard_index (int):
```
dizcology, line 33: Since these wrapper classes are "public", let's acknowledge that this change of adding a new required argument (and in other places) is a breaking change in the commit message.

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -26,44 +27,73 @@
...
+        end_page (int):
+            Required. Last page of the classification
+            (for Splitter/Classifier processors). Default is `0` for other processors.
+        normalized_vertices (List[documentai.NormalizedVertex]):
+            Optional. Vertices for bounding box of `Entity`.
     """
-    documentai_entity: documentai.Document.Entity = dataclasses.field(repr=False)
+
+    shard_index: int = dataclasses.field()
+    documentai_entity: dataclasses.InitVar[documentai.Document.Entity]
```
dizcology, line 61: Just to confirm, are we changing all the existing wrapped documentai messages to `InitVar`, and extract information from the those messages in `__post_init__`, but never keeping the whole documentai message as a field?

(Indeed, we should keep this consistent across all wrapper classes - although technically they are no longer "wrappers" in the usual sense.)

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -26,44 +27,73 @@
...
+            self.start_page = int(page_refs[0].page)
+            self.end_page = int(page_refs[-1].page)
 
-    def crop_image(self, documentai_document: documentai.Document):
+            if page_refs[0].bounding_poly:
+                self.normalized_vertices = page_refs[
+                    0
+                ].bounding_poly.normalized_vertices
+
+    def crop_image(self, documentai_document: documentai.Document) -> Image.Image:
```
dizcology, line 96: Update the return type to `Optional[Image.Image]` since the method could return `None`.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -161,7 +161,7 @@ class Line:
             Required. UTF-8 encoded text.
     """
 
-    documentai_line: documentai.Document.Page.Line
+    documentai_line: dataclasses.InitVar[documentai.Document.Page.Line]
     text: str
```
dizcology, line 165: It feels like we only need the `InitVar` `documentai_line` here, and we can extract the text and store that in a `Line.text` field in `__post_init__` (which has access to `documentai_line`)?

That is, perhaps the `text` field should be a dataclass field with `init=False`, so the caller does not have to pass in both `documentai_line` and `text`?

(Similar situation in some other classes that currently do not have `__post_init__`.)

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -365,10 +365,15 @@ class Page:
...
+            the `google.cloud.documentai.Document.Page`.
         documentai_page (google.cloud.documentai.Document.Page):
-            Required. The original google.cloud.documentai.Document.Page object.
-        text: (str):
-            Required. The full text of the Document containing the Page.
+            Required. The original `google.cloud.documentai.Document.Page` object.
+        text (str):
+            Required. The full text of the `Document` containing the `Page`.
+        page_number (int):
+            Required. The page number of the `Page`.
```
dizcology, line 376: Maybe add to the docstring that `page_number` is always page index+1.  But then how is this field used?  Wouldn't `page_index` be more useful?

(In `samples/snippets/table_sample.py` below, for example, the name `page_number` indeed refers to page index.)

samples/snippets/table_sample.py
```diff
@@ -35,7 +35,7 @@ def table_sample(document_path: str, output_file_prefix: str) -> None:
             # Write table to CSV file
             output_file = f"{output_file_prefix}-{page_number}-{table_number}.csv"
```
dizcology, line 36: Maybe rename these `*_number` to `*_index`.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -131,7 +131,7 @@ class Block:
             Required. UTF-8 encoded text.
     """
 
-    documentai_block: documentai.Document.Page.Block
+    documentai_block: dataclasses.InitVar[documentai.Document.Page.Block]
```
dizcology, line 134: Another small inconsistency is that now only some of the wrapper classes require the `shard_index` init argument, but not all.  Can we make it easy for the next engineer to tell how to decide whether a wrapper class should vs should not have `shard_index`?

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: It should be the same, unless there are any missing (which is a bigger issue tbh)

Maybe I should add an exception for missing shards, because that would cause bigger issues down the line

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: I'm good with that. This was just moving the code that was already present.

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: That is true. I am changing the documentai messages to InitVar for all wrappers except the `Document` wrapper. The `documentai_document` contains all of the information used to populate the other wrappers.

samples/snippets/table_sample.py
holtskinner: Ah, this should be changed to use the actual `Page.page_number` instead of a locally created index.

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: Page number is populated by document ai directly, and it functions as a 1-based index for the page in the document.

I could also add methods to get the original pages/entities by their unique identifiers (page_number and id respectively)

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: I agree

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: Removed these

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -582,7 +559,7 @@ def get_form_field_by_name(self, target_field: str) -> List[FormField]:
         return found_fields
 
     def form_fields_to_dict(self) -> Dict:
-        r"""Returns Dictionary of form fields in document.
+        r"""Returns dictionary of form fields in document.
```
dizcology, line 562: Make the type annotation (and maybe the docstring too) more informative.  Is the return type `Dict[str, str]` here?

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -767,13 +752,15 @@ def export_images(
...
         for entity in self.entities:
-            if not entity.image:
+            image = entity.crop_image(
+                documentai_page=self.pages[entity.start_page].documentai_page
+            )
+            if not image:
                 continue
-
             output_filename = (
                 f"{output_file_prefix}_{index}_{entity.type_}.{output_file_extension}"
```
dizcology, line 761: Update the docstring to include the output filename format.

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -722,27 +699,35 @@ def split_pdf(self, pdf_path: str, output_path: str) -> List[str]:
...
             str:
-                JSON string of TextAnnotations.
+                JSON string of `TextAnnotations`.
         """
         return AnnotateFileResponse.to_json(
-            _convert_to_vision_annotate_file_response(self.text, self.pages)
+            self.convert_document_to_annotate_file_response()
         )
 
     def export_images(
```
dizcology, line 733: The method name here seems to suggest that it exports all images, but actually only those that belong to entities?

(Perhaps all images belong to entities?)

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -26,31 +27,46 @@
...
-    r"""Represents a wrapped documentai.Document.Entity.
+    """Represents a wrapped `documentai.Document.Entity`.
 
     Attributes:
         documentai_entity (google.cloud.documentai.Document.Entity):
-            Required. The original google.cloud.documentai.Document.Entity object.
+            Required. The original `google.cloud.documentai.Document.Entity` object.
+        page_offset (int):
+            Optional. The page offset (for sharded Documents) of the `google.cloud.documentai.Document` containing
+            the `google.cloud.documentai.Document.Entity`. `page_refs.page` is relative to the shard, not the entire Document.
```
dizcology, line 37: `page_refs` seems to be an attribute of the underlying documentai message and is used only in the implementation, and this part of this docstring would not help the user understand what `page_offset` means.

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -59,32 +75,36 @@ def __post_init__(self):
         ):
             self.normalized_text = self.documentai_entity.normalized_value.text
 
-        if self.documentai_entity.page_anchor.page_refs:
-            self.start_page = int(self.documentai_entity.page_anchor.page_refs[0].page)
-            self.end_page = int(self.documentai_entity.page_anchor.page_refs[-1].page)
+        page_refs = self.documentai_entity.page_anchor.page_refs
+        if page_refs:
+            self.start_page = int(page_refs[0].page) + page_offset
+            self.end_page = int(page_refs[-1].page) + page_offset
```
dizcology, line 81: We should add a unit test for these lines of code above.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -29,15 +29,27 @@ class Table:
     Attributes:
         documentai_table (google.cloud.documentai.Document.Page.Table):
             Required. The original google.cloud.documentai.Document.Page.Table object.
+        document_text (str):
+            Required. UTF-8 encoded text in reading order from the document.
```
dizcology, line 33: Is `document_text` the text for the whole document, or just that of a page containing the object (e.g. a Table)?  If it is just the containing page's text, perhaps let's replace the `document_text` attribute with a `_page` attribute instead (here and below).

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: Throughout this code, `document_text` is the text of the full documentai Document, not just the page.

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: Added a more explicit type annotation `Dict[str, Union[str, List[str]]]`

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: This is already in the docstring for the function. Am I missing something?

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: Note - DocAI doesn't do any general image recognition currently, it's only for specific entity extraction

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: Rewrote this doc string. Is this more clear?

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: Done

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: No longer applicable

dizcology (approved):

PR #111, opened 2023-04-26
feat: added hOCR exporter

galz10:


google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+    documentai_text: List[documentai.Document.Page] = dataclasses.field(repr=False)
+    filename: str = dataclasses.field(repr=False)
+
+    hocr_pages: List[hOCR_Page] = dataclasses.field(init=False, repr=False)
+    metadata: str = dataclasses.field(init=False, repr=False)
+
+    def __post_init__(self):
+        self.hocr_pages = _load_pages(pages=(self.documentai_pages), document_text=(self.documentai_text))
+
+    def export_hocr(cls):
```
dizcology, line 209: Why `cls`?  If this is meant to be classmethod, decorate it as so.

There is also no good way to unit test parts of this function as written.

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
+# uncompyle6 version 3.9.0
```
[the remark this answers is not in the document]
galz10: I was unaware of the dataclass style, i fixed those and removed the uncomplye6 stuff

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+from google.cloud import documentai
+
+ElementWithLayout = Union[(
+ documentai.Document.Page.Paragraph,
+ documentai.Document.Page,
+ documentai.Document.Page.Token,
+ documentai.Document.Page.Block,
+ documentai.Document.Page.Line)]
+
+def _get_bounding_box(object: ElementWithLayout, dimensions: documentai.Document.Page.Dimension):
```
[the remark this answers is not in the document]
galz10: done

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+    start_index = object.layout.text_anchor.text_segments[0].start_index
+    end_index = object.layout.text_anchor.text_segments[0].end_index
+    return document_text[start_index:end_index].replace('/n', '')
+
+
+def _load_hocr_words(words: List[documentai.Document.Page.Token], document_text: str, dimensions: documentai.Document.Page.Dimension):
+    w = []
+    for word in words:
+        text = _get_text(word, document_text)
+        w.append(hOCR_Word(text=text, documentai_word=word, dimensions=dimensions))
```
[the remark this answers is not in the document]
galz10: If the object only has normalized vertices it uses the normalized vertices and the dimensions to convert to pixels.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+        return p
+
+
+def _load_hocr_blocks(page: documentai.Document.Page, document_text: str):
+    blocks = []
+    for block in page.blocks:
+        start_index = block.layout.text_anchor.text_segments[0].start_index
+        end_index = block.layout.text_anchor.text_segments[0].end_index
+        paragraphs = [p for p in page.paragraphs if p.layout.text_anchor.text_segments[0].start_index >= start_index if p.layout.text_anchor.text_segments[0].end_index <= end_index]
+        blocks.append(hOCR_Block(paragraphs=paragraphs, page=page, dimensions=(page.dimension), id=1, document_text=document_text, documentai_block=block))
```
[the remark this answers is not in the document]
galz10: So i ended up not using `id` i removed it.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+    else:
+        return w
+
+
+def _load_hocr_lines(lines: List[documentai.Document.Page.Line], page: documentai.Document.Page, document_text: str):
+    l = []
+    for line in lines:
+        start_index = line.layout.text_anchor.text_segments[0].start_index
+        end_index = line.layout.text_anchor.text_segments[0].end_index
+        words = [word for word in page.tokens if word.layout.text_anchor.text_segments[0].start_index >= start_index if word.layout.text_anchor.text_segments[0].end_index <= end_index]
```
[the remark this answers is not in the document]
galz10: If we put the `_load_hocr_lines,_load_hocr_words,etc...` logic in the respective toolbox wrapper, it would make instantiation super slow. The approach 2 implementation is really slow compared to this implementation.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+def _load_pages(pages: List[documentai.Document.Page], document_text: str):
+    res_pages = []
+    for page in pages:
+        res_pages.append(hOCR_Page(documentai_page=(page.documentai_page), document_text=document_text))
+    else:
+        return res_pages
+
+
+@dataclasses.dataclass
+class hOCR_Word:
```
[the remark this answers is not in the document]
galz10: Done

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: You could mock the respective objects and assert the returned string, and you can then test the respective objects _load functions.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+        f += "<meta name=\"ocr-capabilities\" content=\"ocr_page ocr_carea ocr_par ocr_line ocrx_word\" />\n"
+        f += "</head>\n"
+        f += "<body>\n"
+        for pidx,page in enumerate(cls.hocr_pages):
+            f += f"<div class='ocr_page' lang='unknown' title='image \"{cls.filename}\";{page.bounding_box}'>\n"
+            for bidx,block in enumerate(page.hocr_blocks):
+                f += f"<span class='ocr_carea' id='block_{pidx}_{bidx}' title='{block.bounding_box}'>\n"
+                for paridx, paragraph in enumerate(block.hocr_paragraphs):
+                    f += f"<span class='ocr_par' id='par_{pidx}_{bidx}_{paridx}' title='{paragraph.bounding_box}'>\n"
+                    for lidx, line in enumerate(paragraph.hocr_lines):
```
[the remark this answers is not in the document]
galz10: Done.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+    if object.layout.bounding_poly.vertices:
+        min_x, min_y = object.layout.bounding_poly.vertices[0].x, object.layout.bounding_poly.vertices[0].y
+        max_x, max_y = object.layout.bounding_poly.vertices[2].x, object.layout.bounding_poly.vertices[2].y
+        return f"bbox {int(min_x)} {int(min_y)} {int(max_x)} {int(max_y)}"
+    min_x, min_y = object.layout.bounding_poly.normalized_vertices[0].x, object.layout.bounding_poly.normalized_vertices[0].y
+    max_x, max_y = object.layout.bounding_poly.normalized_vertices[2].x, object.layout.bounding_poly.normalized_vertices[2].y
+    return f"bbox {int(min_x * dimensions.width)} {int(min_y * dimensions.height)} {int(max_x * dimensions.width)} {int(max_y * dimensions.height)}"
+
+
+def _get_text(object: ElementWithLayout, document_text):
```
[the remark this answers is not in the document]
galz10: _get_text and _get_bounding box is only used for hOCR it's not used anywhere else so i don't know if there would be a benefit of creating a ElementWithLayout class. Would making a ElementWithLayout class also mess with using document ai objects ?

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,241 @@
...
+ documentai.Document.Page,
+ documentai.Document.Page.Token,
+ documentai.Document.Page.Block,
+ documentai.Document.Page.Line)]
+
+def _get_bounding_box(object: ElementWithLayout, dimensions: documentai.Document.Page.Dimension):
+    if object.layout.bounding_poly.vertices:
+        min_x, min_y = object.layout.bounding_poly.vertices[0].x, object.layout.bounding_poly.vertices[0].y
+        max_x, max_y = object.layout.bounding_poly.vertices[2].x, object.layout.bounding_poly.vertices[2].y
+        return f"bbox {int(min_x)} {int(min_y)} {int(max_x)} {int(max_y)}"
```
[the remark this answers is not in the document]
galz10: are you suggesting a `bounding_poly` wrapper just for hOCR or for all of docai toolbox.

holtskinner:
In the description for this PR, (and the inline documentation) can you add context as to what hOCR is and how it's being used?

galz10:
> In the description for this PR, (and the inline documentation) can you add context as to what hOCR is and how it's being used?

Yes, will do this is just a draft to help finalize the design doc but when i make this not a draft i'll add the context.

google/cloud/documentai_toolbox/wrappers/hocr.py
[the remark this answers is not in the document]
dizcology: For Document AI's bounding poly messages: https://github.com/googleapis/googleapis/blob/57b675c1534e6feb806ca9cb48a3c4a4023e91fe/google/cloud/documentai/v1/geometry.proto#L49.  Alternatively, at least we should extract some of the code here to helper functions for simpler testing.

google/cloud/documentai_toolbox/wrappers/hocr.py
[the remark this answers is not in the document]
dizcology: Good point, in this case perhaps have them as hOCR-specific private helpers.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
+import dataclasses
+from typing import List, Union
+from google.cloud import documentai
+
+from google.cloud.documentai_toolbox.constants import ElementWithLayout
+
+
+def _get_bounding_box(element_with_layout: ElementWithLayout, dimensions: documentai.Document.Page.Dimension):
+    if element_with_layout.layout.bounding_poly.vertices:
```
dizcology, line 9: Looks like there are `vertices` and there are `normalized_vertices`.  What are the situations where only `normalized_vertices` are populated but not `vertices`?  (That looks like the only case where `dimensions` is used, and right not we pass this to the hOCR classes just for this purpose.  This perhaps could be avoided.)

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+        res_pages.append(_HocrPage(documentai_page=page.documentai_page, document_text=document_text, filename=filename))
+
+    return res_pages
+
+
+@dataclasses.dataclass
+class _HocrWord:
+    text: str = dataclasses.field(repr=False)
+    documentai_word: documentai.Document.Page.Token = dataclasses.field(repr=False)
+    dimensions: documentai.Document.Page.Dimension = dataclasses.field(repr=False)
```
dizcology, line 78: This might be an `InitVar` following the pattern introduced in https://github.com/googleapis/python-documentai-toolbox/pull/110

(The word does not need this as an attribute, and it is used only in `__post_init__` to calculate the bounding box.)

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+    for page in pages:
+        res_pages.append(_HocrPage(documentai_page=page.documentai_page, document_text=document_text, filename=filename))
+
+    return res_pages
+
+
+@dataclasses.dataclass
+class _HocrWord:
+    text: str = dataclasses.field(repr=False)
+    documentai_word: documentai.Document.Page.Token = dataclasses.field(repr=False)
```
dizcology, line 77: Also this, perhaps just an `InitVar` since the hOCR word does not need it as an attribute.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+    dimensions: documentai.Document.Page.Dimension = dataclasses.field(repr=False)
+    bounding_box: str = dataclasses.field(init=False, repr=False)
+
+    def __post_init__(self):
+        self.bounding_box = _get_bounding_box(element_with_layout=(self.documentai_word), dimensions=(self.dimensions))
+
+    def to_hocr(self, pidx: int,bidx: int, paridx: int, lidx: int, widx: int):
+        f = ""
+        word_text = self.text.replace("&","&amp;").replace("<","&lt;").replace(">","&gt;")
```
dizcology, line 86: There must be library for this.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+        f += f"<span class='ocrx_word' id='word_{pidx}_{bidx}_{paridx}_{lidx}_{widx}' title='{self.bounding_box}'>{word_text}</span>\n"
```
dizcology, line 87: Similarly, there are standard libraries for manipulating and creating XML: https://docs.python.org/3/library/xml.etree.elementtree.html

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -760,3 +762,8 @@ def export_images(
             index += 1
 
         return output_filenames
+
+    def to_hocr(self,filename: str) -> str:
+        hocr = _Hocr(documentai_pages=self.pages, documentai_text=self.text,filename=filename)
+        
+        return hocr.export_hocr()
```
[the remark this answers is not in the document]
dizcology: To be clear, I am not requesting to change this at this point, let's continue the discussion.

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+        f += "<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"unknown\" lang=\"unknown\">\n"
+        f += "<head>\n"
+        f += f"<title>{self.filename}</title>\n"
+        f += "<meta http-equiv=\"Content-Type\" content=\"text/html;charset=utf-8\" />\n"
+        f += "<meta name=\"ocr-system\" content=\"Document AI OCR\" />\n"
+        f += "<meta name=\"ocr-langs\" content=\"unknown\" />\n"
+        f += f"<meta name=\"ocr-number-of-pages\" content=\"{len(self.documentai_pages)}\" />\n"
+        f += "<meta name=\"ocr-capabilities\" content=\"ocr_page ocr_carea ocr_par ocr_line ocrx_word\" />\n"
+        f += "</head>\n"
+        f += "<body>\n"
```
dizcology, line 215: Put everything that is constant (does not depend on the actual pages) into a single multiline f-string.

google/cloud/documentai_toolbox/wrappers/hocr.py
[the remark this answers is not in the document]
galz10: okay will do.

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: i've seen cases where the OCR doesn't populate vertices, but i'll test this to validate that it happens sometimes otherwise if it doesn't happen the dimensions is not needed.

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: yea once the #110 changes are merged i'll modify this PR to use the new pattern

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: Same as above

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: to do the text replace stuff or the hOCR stuff? the only thing i found about hOCR was not maintained and did not work properly. For the text replace stuff there might be but wouldn't importing a library just for this purpose be overkill ?

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: oh i did not know i could do this through a library i'll take a look.

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: will do.

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
galz10: I've tried to use only helper functions without the loading object steps and the result is a super slow export, with this implementation it takes milliseconds for the export to return a string and the second implementation it takes seconds up to a minute.

google/cloud/documentai_toolbox/converters/vision_helpers.py
```diff
@@ -24,6 +24,8 @@
 from google.cloud.vision import TextAnnotation, Symbol, Word, Paragraph, Block, Page
 from google.cloud import vision
 
+from google.cloud.documentai_toolbox.constants import ElementWithLayout
```
dizcology, line 27: I noticed that we are not very consistent with importing modules versus importing names/classes.  The general preference is to import modules.  For example the next engineer has to refer to the import line to know if `Page` here is from the Toolbox itself or from `google.cloud.vision`.

(This does not have to be done in this PR, but please add a internal clean up tracking issue to fix this.)

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -48,6 +48,8 @@
 
 from pikepdf import Pdf
 
+from google.cloud.documentai_toolbox.wrappers.hocr import _Hocr
```
dizcology, line 51: Rename to `_hocr` to signal that it is not part of the public API.  Also import the module instead of the class here.

google/cloud/documentai_toolbox/wrappers/hocr.py
dizcology: Can we also confirm this with the service team?

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+        min_x, min_y = element_with_layout.layout.bounding_poly.vertices[0].x, element_with_layout.layout.bounding_poly.vertices[0].y
+        max_x, max_y = element_with_layout.layout.bounding_poly.vertices[2].x, element_with_layout.layout.bounding_poly.vertices[2].y
+        return f"bbox {int(min_x)} {int(min_y)} {int(max_x)} {int(max_y)}"
+    min_x, min_y = element_with_layout.layout.bounding_poly.normalized_vertices[0].x, element_with_layout.layout.bounding_poly.normalized_vertices[0].y
+    max_x, max_y = element_with_layout.layout.bounding_poly.normalized_vertices[2].x, element_with_layout.layout.bounding_poly.normalized_vertices[2].y
+    return f"bbox {int(min_x * dimensions.width)} {int(min_y * dimensions.height)} {int(max_x * dimensions.width)} {int(max_y * dimensions.height)}"
+
+
+def _get_text(element_with_layout: ElementWithLayout, document_text):
+    start_index = element_with_layout.layout.text_anchor.text_segments[0].start_index
```
dizcology, line 19: Similarly here, we should consider either 

(1) promoting `ElementWithLayout` to a proper class (and not just a type alias) so that we can put the code `element_with_layout.layout.text_anchor.text_segments[0].start_index` into an `ElementWithLayout.start_index` property.  (And in fact, then, `_get_text` could be just a property of that class.)

or 

(2) introduce a private wrapper class of `Layout` or `TextAnchor`.

(This refactoring appears to be purely internal and does not have to block the hOCR export work.)

google/cloud/documentai_toolbox/wrappers/hocr.py
```diff
@@ -0,0 +1,221 @@
...
+    end_index = element_with_layout.layout.text_anchor.text_segments[0].end_index
+    return document_text[start_index:end_index].replace('/n', '')
+
+
+def _load_hocr_words(words: List[documentai.Document.Page.Token], document_text: str, dimensions: documentai.Document.Page.Dimension):
```
dizcology, line 24: (1) It may be simpler to test if the method converts a single Document AI Token to a single hOCR Word, and let the caller handle the `for` loop.

(2) The function name could provide even more information, such as `hocr_words_from_documentai_tokens`, or the singular form (if you move the `for` loop out), `hocr_word_from_documentai_token`.

Side note: there are other alternatives that is more object-oriented, now that we have the classes.  For example

```
class _HocrWord:
  ...
  @classmethod
  def from_word(cls, word: wrappers.Word) -> '_HocrWord':
    ...
```
This still requires figuring out whether a "word" needs to know about the page's dimensions (which does not make sense to me), but indeed, the `text` should be already contained in the `word` and having to pass both in seems like a sign of something else not working properly.

google/cloud/documentai_toolbox/wrappers/hocr.py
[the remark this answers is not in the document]
dizcology: The Toolbox is the reasonable place to implement that logic (which we are doing here, but doing it in a way to be used only for hOCR export, while it is a more general concept that belongs to the Document AI Documents themselves).  Should we not expect the user to want to iterate over all the lines on a Toolbox-wrapped Page, and do something about the lines?

google/cloud/documentai_toolbox/wrappers/hocr.py
dizcology: https://stackoverflow.com/questions/1061697/whats-the-easiest-way-to-escape-html-in-python

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
galz10: So I can try to add this logic inside the normal wrappers to see if there is any affect I believe there will be. I believe the reason it would take more time is because we're adding another operation to each token, line , paragraph, block and page wrapping process. So when you have N pages with N tokens,lines,paragraphs and blocks the time added to each wrapping I believe would be a lot(I think the time complexity would be O(n^5). As opposed to only exporting the objects to hOCR when the user actually wants to do export the wrapped document.

google/cloud/documentai_toolbox/wrappers/hocr.py
galz10: Yea that's something we can do to simplify our usage of these across the toolbox, I think it might be beneficial to add this change in a separate PR and not the PR for hOCR feature.

PR #123, opened 2023-06-06
feat: added hOCR export functionality

galz10:
This PR is based off of #111 


google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -411,9 +603,73 @@ def __post_init__(self):
...
+                A string hOCR version of the documentai.Document.Page.
+        """
+        f = ""
+        dimensions = self.documentai_page.dimension
+        pidx = self.documentai_page.page_number
+        page_bounding_box = _get_hocr_bounding_box(
+            element_with_layout=(self.documentai_page), dimensions=(dimensions)
+        )
+        f += f"<div class='ocr_page' lang='unknown' title='image \"{filename}\";{page_bounding_box}'>\n"
+        for bidx, block in enumerate(self.blocks):
```
dizcology, line 639: Instead of nested for-loops, is it feasible to have an `to_hocr` method in the wrapper classes Page, Block, Line, Token and call it recursively?  (This may be an abstract method in a parent base class for all wrapper classes that support `to_hocr`.) 

So the method body becomes something similar to:

```
class _HocrBase:
  @abstractmethod
  @property
  def children(self) -> List[Any]:
    ...

  def to_hocr(self, filename):  # ideally we don't have to ask for filename
    result = "..."  # page-level info
    for i, child in self.children:
      result += child.to_hocr(filename)
    result += "</span>"  # page-level.
```

I think this can be considered a follow-up refactoring work and does not need to be done in this PR.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -411,9 +603,73 @@ def __post_init__(self):
...
+                        )
+                        word_text = (
+                            token.text.replace("&", "&amp;")
+                            .replace("<", "&lt;")
+                            .replace(">", "&gt;")
+                        )
+                        f += f"<span class='ocrx_word' id='word_{pidx}_{bidx}_{paridx}_{lidx}_{tidx}' title='{token_bounding_box}'>{word_text}</span>\n"
+                f += "</span>\n"
+            f += "</span>\n"
+        f += "</div>\n"
```
dizcology, line 674: I wonder if we can omit these newlines, and at the end optionally call an xml-formatter (if the user wants the xml string to look pretty).

galz10: yea the newlines can be omitted the looking pretty doesn't matter to the initial customer

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -206,8 +263,96 @@ def _text_from_layout(layout: documentai.Document.Page.Layout, text: str) -> str
     return result_text
 
 
-def _get_blocks(blocks: List[documentai.Document.Page.Block], text: str) -> List[Block]:
-    r"""Returns a list of Block.
+def _get_tokens_in_line(
+    line: documentai.Document.Page.Line, tokens: List[Token]
```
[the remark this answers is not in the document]
galz10: the return type is different for them, `_get_tokens_in_line` returns List[Token], `_get_lines_in_paragraph` returns List[Line], ` _get_paragraphs_in_blocks` return List[Paragraph]

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,49 +92,53 @@ def sample_table_to_csv():
...
+            Required. The original google.cloud.documentai.Document.Page.FormField object.
+        field_name (str):
+            Required. The form field name
+        field_value (str):
+            Required. The form field value
+    """
+
+    documentai_formfield: documentai.Document.Page.FormField
+    field_name: str
+    field_value: str
```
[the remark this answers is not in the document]
galz10: I don't know why this is showing up as a new change but this i did not add this, this is from Holts addition. The variables `field_name` and `field_value` and meant to make it easier to access similar to `entity.type_` and `entity.mention_text`

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,49 +92,53 @@ def sample_table_to_csv():
...
+        documentai_line (google.cloud.documentai.Document.Page.Line):
+            Required. The original google.cloud.documentai.Document.Page.Line object.
         text (str):
             Required. UTF-8 encoded text.
     """
 
-    documentai_block: documentai.Document.Page.Block
+    documentai_line: documentai.Document.Page.Line
     text: str
+    tokens: List[Token]
```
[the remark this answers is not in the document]
galz10: yea users should never wrap their own `Line,Entity,Paragraph,etc` they should only use the Document.from_ to wrap a whole document not specific parts of a document.  My thought process if they are lazy calculated we'll need to pass in the whole page to be able to access things. I think for the initial release this might be okay and if there are performance issues we can go and refactor to do lazy loading.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: i think for the purpose of releasing this and letting the customer be able to use this, we should skip this and add it to backlog for refactoring

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -765,3 +765,33 @@ def export_images(
             index += 1
 
         return output_filenames
+
+    def export_hocr_str(self, filename: str) -> str:
+        r"""Exports a string hOCR version of the Document.
+
+        Args:
+            filename (str):
+                Required. The filename of the original file.
```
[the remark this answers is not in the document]
galz10: You need the filename for the hOCR, i don't believe the DocAI response has any filename or something close to it so that's why we ask the caller for that information.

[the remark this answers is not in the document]
dizcology: In the current code, `filename` is used for `title`, is it the "title" of the whole hOCR document?  If so perhaps we should just call it `title`?  Also if it belongs only to the top level (the whole hOCR document), perhaps it does not need to be passed to the lower level elements (e.g. hOCR page).

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
dizcology: Sounds good on not worrying about lazy calculation until later.

On the other hand, let's evaluate whether the whole page should be passed to every element as a private attribute.  Only a reference of the page will be passed so there will be no copies.  As is right now, every time an element with layout needs to look up some text, the caller of the method has to pass in a page, since only the page has the text (is this true?).

There seems no harm to just include the page of an element (Block, Line, etc) as a private attribute of the element, and it would simplify the code (we will no longer need to pass in `text` in the methods).

This is assuming that every element belongs to only one page, though.  Do we know this is true?

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
dizcology: Did you perhaps merge/rebase with Holt's PR at some point?  It seems like we will encounter a pretty bad merge conflict if this is not cleaned up very soon.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -148,39 +155,90 @@ class Paragraph:
...
+        return f"bbox {int(min_x)} {int(min_y)} {int(max_x)} {int(max_y)}"
+    min_x, min_y = (
+        element_with_layout.layout.bounding_poly.normalized_vertices[0].x,
+        element_with_layout.layout.bounding_poly.normalized_vertices[0].y,
+    )
+    max_x, max_y = (
+        element_with_layout.layout.bounding_poly.normalized_vertices[2].x,
+        element_with_layout.layout.bounding_poly.normalized_vertices[2].y,
+    )
+    return f"bbox {int(min_x * dimension.width)} {int(min_y * dimension.height)} {int(max_x * dimension.width)} {int(max_y * dimension.height)}"
```
dizcology, line 241: I think the "switch" nature is more obvious if we put second half of the code in an `else` statement.  Also the code here is largely duplicated, why not refactor this out into a little function that can be easily tested, for example

```
def _xy_from_vertices(vertices: ..., min: bool = True) -> Tuple[int, int]:
    ... 
```

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
dizcology: I think there are fancy ways to properly express that with TypeVar, but we probably don't gain much doing that.  Is there anything wrong to return a list of `ElementWithLayout`, which is a union type?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -206,8 +264,96 @@ def _text_from_layout(layout: documentai.Document.Page.Layout, text: str) -> str
...
+        if token.documentai_token.layout.text_anchor.text_segments[0].start_index
+        >= start_index
+        if token.documentai_token.layout.text_anchor.text_segments[0].end_index
+        <= end_index
+    ]
+
+
+def _get_lines_in_paragraph(
+    paragraph: documentai.Document.Page.Paragraph, lines: List[Line]
+) -> List[Paragraph]:
```
dizcology, line 297: This annotation of return type seems incorrect.

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Agreed.  Let's add an issue of low-ish priority to track this.  This kind of refactoring makes things easier to test locally (we can test each level separately), and when things break it will be much easier to pinpoint where things break since the tests would be smaller.

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
galz10: No this was pulled from Main, not holts PR but once holt's PR is submitted i will need to change some of the implementation probably

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
galz10: So it's not returning the Wrapped version of `Lines,Paragraph,etc...` so we wouldn't be able to use `ElementWithLayout` but we can probably make a similar type, would each type need to have the same objects for example `Paragraph` has similar variables except `lines` and `Line` has `tokens` , would that still be fine?

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
galz10: So it every element that is loaded is associated to a specific page so as long as we pass in the page i think we will be fine. I'll try out passing the page as reference.

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
galz10: Will do, i don't think filename gets passed down to lower elements.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: wil do.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: Correct.

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
galz10: Just checked, ocr_page needs a title also so that's why it's passed down to page

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -16,11 +16,27 @@
...
-from typing import List
+import html
+from typing import List, TypeVar, Union
+
+from google.cloud.documentai_toolbox.constants import ElementWithLayout
 
 from google.cloud import documentai
 import pandas as pd
 
+Page = TypeVar("Page")
```
dizcology, line 27: Let's not use TypeVar for this.  In the annotations below, you can simply quote the name of the type.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
-    header_rows = _table_rows_from_documentai_table_rows(
-        table_rows=list(documentai_table.header_rows), text=text
-    )
-    body_rows = _table_rows_from_documentai_table_rows(
-        table_rows=list(documentai_table.body_rows), text=text
+    documentai_object: documentai.Document.Page.Line = dataclasses.field(
+        init=True, repr=False
     )
+    text: str = dataclasses.field(init=True, repr=False)
+    _page: Page = dataclasses.field(init=True, repr=False)
```
dizcology, line 139: Here the annotation can be: (note the quotes)
```
_page: "Page" = ...
```

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
 
-    header_rows = _table_rows_from_documentai_table_rows(
-        table_rows=list(documentai_table.header_rows), text=text
-    )
-    body_rows = _table_rows_from_documentai_table_rows(
-        table_rows=list(documentai_table.body_rows), text=text
+    documentai_object: documentai.Document.Page.Line = dataclasses.field(
+        init=True, repr=False
     )
+    text: str = dataclasses.field(init=True, repr=False)
```
dizcology, line 138: If we already pass in the page, we should not need the `text` field anymore.  If we are concerned about removing `text` as a breaking change, we can add a `text` property that returns `self._page.text`.

(Since there are other breaking changes introduced in this PR, I don't think that is very important.)

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
+
+@dataclasses.dataclass
+class Line:
+    """Represents a wrapped documentai.Document.Page.Line.
+
+    Attributes:
+        documentai_object (google.cloud.documentai.Document.Page.Line):
+            Required. The original google.cloud.documentai.Document.Page.Line object.
+        text (str):
+            Required. UTF-8 encoded text.
```
dizcology, line 132: Update docstring about the new `_page` field.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
 
-    documentai_paragraph: documentai.Document.Page.Paragraph
-    text: str
+    documentai_object: documentai.Document.Page.Paragraph = dataclasses.field(
+        init=True, repr=False
+    )
+    text: str = dataclasses.field(init=True, repr=False)
+    _page: Page = dataclasses.field(init=True, repr=False)
+
+    lines: List[Line] = dataclasses.field(init=False, repr=False)
```
dizcology, line 186: Just a reminder: let's be sure to have one unit test for every element that verifies these children element fields are correctly populated.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
+    )
+
+    return Table(
+        documentai_table=documentai_table, body_rows=body_rows, header_rows=header_rows
+    )
+
+
+def _get_hocr_bounding_box(
+    element_with_layout: ElementWithLayout,
+    dimension: documentai.Document.Page.Dimension,
```
dizcology, line 249: We may not need this any more, since every element_with_layout has access to its containing page?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +106,184 @@ def sample_table_to_csv():
...
+        dimension (documentai.Document.Page.Dimension):
+            Required. Page dimension.
+
+    Returns:
+        str:
+            hOCR bounding box sring.
+    """
+
+    if element_with_layout.layout.bounding_poly.vertices:
+        min_x, min_y = (
```
dizcology, line 265: Let's refactor this duplicate code (with the code below) into something that can be easily unit-tested just once.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -217,25 +348,28 @@ def _get_blocks(blocks: List[documentai.Document.Page.Block], text: str) -> List
...
             )
         )
 
     return result
 
 
 def _get_paragraphs(
-    paragraphs: List[documentai.Document.Page.Paragraph], text: str
+    paragraphs: List[documentai.Document.Page.Paragraph],
+    text: str,
```
dizcology, line 369: For these helper functions too, we should not need to have a `text` argument any more.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -245,23 +379,53 @@ def _get_paragraphs(
...
+                documentai_object=paragraph,
                 text=_text_from_layout(layout=paragraph.layout, text=text),
+                _page=_page,
+            )
+        )
+
+    return result
+
+
+def _get_tokens(tokens: List[documentai.Document.Page.Token], text: str) -> List[Token]:
```
dizcology, line 398: Should this get a `page` instead, like the other helper functions?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -217,25 +348,28 @@ def _get_blocks(blocks: List[documentai.Document.Page.Block], text: str) -> List
...
         )
 
     return result
 
 
 def _get_paragraphs(
-    paragraphs: List[documentai.Document.Page.Paragraph], text: str
+    paragraphs: List[documentai.Document.Page.Paragraph],
+    text: str,
+    _page: Page,
```
dizcology, line 370: As an argument of a private helper function, let's not prefix the argument name with underscore.  Plain `page` for the argument name is fine.

(For the fields/attributes, we prefix with underscore to signal to the user that that field is private and since this is Python the user is free to access it, but we make no guarantees that API is table.)

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -400,7 +566,13 @@ class Page:
...
         tables = []
-
+        """
+        Order of Init
+        Token
+        Line
+        Paragraph,
+        Block
+        """
         for table in self.documentai_page.tables:
```
dizcology, line 576: If we are going to rename `documentai_line` to `documentai_object` as you have done in other places, let's keep this consistent and change it for the `Page` class too.

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Also by the way I believe `init=True` is the default, so it does not need to be explicitly passed to `dataclasses.field`.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: How can we refactor this more? We're dealing with two separate objects `vertices` and `normalized vertices`

google/cloud/documentai_toolbox/wrappers/page.py
galz10: it's a little difficult to test the `Lines,Tokens etc` by themselves due to the `_page` addition but i did expand `test_Page` to check children

google/cloud/documentai_toolbox/wrappers/page.py
galz10: we will still need it since we need to get the bounding boxes, we can instead of calling this in `to_hocr` we can call it at object initialization.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: No since this doesn't have children so there is no need to pass in a page.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: text is referring to the text in the line not the text in the page

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -90,97 +100,188 @@ def sample_table_to_csv():
...
-            Required. UTF-8 encoded text in reading order
-            from the document.
+            Required. UTF-8 encoded text.
+    """
 
-    Returns:
-        Table:
-            A Table.
+    documentai_object: documentai.Document.Page.Token
+    text: str
```
dizcology, line 115: Maybe change this to `_page` too?

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Sounds good.

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Is the page's dimension already included in the element's `_page`?

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Maybe something like the following?

```
def _get_xy(..., normalized: bool=False, min: bool=False):
    ...
```

To be sure, such a function will still need to be tested for all different combinations of use cases (`normalized=True`, `normalized=False` etc) perhaps in a parametrized test.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -206,8 +307,36 @@ def _text_from_layout(layout: documentai.Document.Page.Layout, text: str) -> str
     return result_text
 
 
-def _get_blocks(blocks: List[documentai.Document.Page.Block], text: str) -> List[Block]:
-    r"""Returns a list of Block.
+def _get_children_of_element(element: ElementWithLayout, children: ChildrenElements):
+    r"""Returns a list of paragraph inside block.
+
+    Args:
+        block (documentai.Document.Page.Block):
```
dizcology, line 314: Update the docstring.

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: For consistency perhaps let's use `page` here instead of text.  I understand this is passing in more information than needed, but this removes a special case among a family of helper functions.

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: Got it - can we make that a cached property (instead of asking the caller of the constructor to do the calculation), since this is now easily computable from `documentai_object` and `_page`?

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +612,76 @@ class Page:
...
+        )
+        self.page_number = int(self.documentai_object.page_number)
+        tables = []
+        """
+        Order of Init
+        Token
+        Line
+        Paragraph,
+        Block
+        """
```
holtskinner, line 629: Move this comment to the top of the function

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +612,76 @@ class Page:
...
+        self.tokens = _get_tokens(tokens=self.documentai_object.tokens, page=self)
+        self.lines = _get_lines(lines=self.documentai_object.lines, page=self)
+        self.paragraphs = _get_paragraphs(
+            paragraphs=self.documentai_object.paragraphs, page=self
+        )
+        self.blocks = _get_blocks(
+            blocks=self.documentai_object.blocks,
+            page=self,
+        )
+        self.tables = tables
```
holtskinner, line 649: Remove

holtskinner (changes requested):
General comment, is there a reason all of the hocr stuff is in the document and page classes instead of a converter class/file like the third party and vision converters? I understand the refactoring needed to the Page classes to make this work, but I feel like the hocr stuff should be brought out to keep the wrapper classes more single-purpose. What do you think?

holtskinner:
Maybe the owlbot formatting messed with the readme gen? I'm not sure why it decided to change the quote style in that file.

tests/unit/test_page.py
```diff
@@ -133,39 +133,81 @@ def test_trim_text():
...
+        documentai_object=docproto.pages[0], document_text=docproto.text
     )
-    assert body_rows == [
-        ["Tool A", "500", "$1.00", "$500.00"],
-        ["Service B", "1", "$900.00", "$900.00"],
-        ["Resource C", "50", "$12.00", "$600.00"],
-        ["", "", "Subtotal", "$2000.00"],
-        ["", "", "Tax", "$140.00"],
-        ["", "", "BALANCE DUE", "$2140.00"],
-    ]
```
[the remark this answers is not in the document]
galz10: `_body_for_table_rows_from_documentai_table_rows` is not being used anywhere

tests/unit/test_page.py
```diff
@@ -133,39 +133,81 @@ def test_trim_text():
...
-    assert text == "Invoice\n"
+    paragraphs = page._get_paragraphs(paragraphs=docproto_paragraphs, page=wrapped_page)
+
+    assert len(paragraphs) == 31
+    assert paragraphs[0].text == "Invoice\n"
+    assert paragraphs[0].hocr_bounding_box == "bbox 1310 220 1534 282"
+    # checking cached value
+    assert paragraphs[0].text == "Invoice\n"
+    assert paragraphs[0].hocr_bounding_box == "bbox 1310 220 1534 282"
+
```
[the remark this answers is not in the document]
galz10: All `_get_*` does is do the same thing you are doing for formfields just in a separate function to make it more readable.

tests/unit/test_document.py
```diff
@@ -637,9 +640,22 @@ def test_export_images(get_bytes_images_mock):
...
 
-    assert os.path.exists(output_path)
-    shutil.rmtree(output_path)
-
     assert actual == [
         "exported_photo_0_Portrait.png",
     ]
+
+    assert os.path.exists(output_path)
+    shutil.rmtree(output_path)
```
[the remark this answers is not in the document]
galz10: i tried adding this in a fixture but it was not working properly

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -103,80 +113,91 @@ def sample_table_to_csv():
...
+    def text(self):
+        if self._text is None:
+            self._text = _text_from_layout(
+                layout=self.documentai_object.layout, text=self._page.text
+            )
+        return self._text
+
+    @property
+    def hocr_bounding_box(self):
+        print(self._hocr_bounding_box)
```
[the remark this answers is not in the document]
galz10: nope good catch.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -16,18 +16,28 @@
...
+from google.cloud.documentai_toolbox.constants import ElementWithLayout
 
 from google.cloud import documentai
 import pandas as pd
 
+ChildrenElements = Union[
+    List["Paragraph"],
+    List["Block"],
+    List["Token"],
+    List["Line"],
```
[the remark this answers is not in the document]
galz10: Since we are referencing something that is inside the file you can't use `Block,Paragraph...etc` i had it with TypeVar so you could use `Block,Paragraph...etc` but Yu-han suggested to just use the quotes.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -103,80 +113,91 @@ def sample_table_to_csv():
...
-    document_text: dataclasses.InitVar[str]
-    text: str = dataclasses.field(init=False)
-
-    def __post_init__(self, document_text) -> None:
-        self.text = _text_from_layout(
-            layout=self.documentai_block.layout, text=document_text
-        )
-
+    documentai_object: documentai.Document.Page.Token = dataclasses.field(default=None)
+    _page: "Page" = dataclasses.field(default=None)
```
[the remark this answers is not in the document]
galz10: Yes, the other way is using TypeVar but it requires an import of another library. https://stackoverflow.com/questions/72039261/type-hint-to-return-an-instance-of-a-class-where-typevar-is-the-class-type

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -267,13 +333,249 @@ def _table_rows_from_documentai_table_rows(
...
+    min: bool = False,
+):
+    r"""Returns xy coordinates corresponding to element.
+
+    Args:
+        element (ElementWithLayout):
+            Required. an element with layout fields.
+        dimension (documentai.Document.Page.Dimension):
+            Required. Page dimension.
+        normalized (Boolean):
```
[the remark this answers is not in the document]
galz10: the get_xy takes in an element with layout so you must pass either `vertices` or` vertices` you can't pass the `bounding_poly` object and check. Also this is trying to reduce duplicate code having a function that you can pass in a ElementWithLayout and a couple variables reduces duplicate code.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -267,13 +333,249 @@ def _table_rows_from_documentai_table_rows(
...
+
+    for token in tokens:
+        result.append(
+            Token(
+                documentai_object=token,
+                _page=page,
+            )
+        )
+
+    return result
```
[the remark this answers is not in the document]
galz10: All these `_get_*` functions corresponding to `Line,Token,Paragraph,Block` they do the same thing you're doing [here](https://github.com/googleapis/python-documentai-toolbox/blob/main/google/cloud/documentai_toolbox/wrappers/page.py#L317) the only difference is making the `post_init` of page easier to read.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +612,76 @@ class Page:
...
+        return f
+
+    @property
+    def hocr_bounding_box(self):
+        if self._hocr_bounding_box is None:
+            self._hocr_bounding_box = _get_hocr_bounding_box(
+                element_with_layout=self.documentai_object,
+                dimension=self.documentai_object.dimension,
+            )
+        return self._hocr_bounding_box
```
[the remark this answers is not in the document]
galz10: The hOCR needs to have all the information of the page and unlike all the other classes the hOCR exporter will not have any additions or features since it's only purpose is for exporting.

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -767,3 +767,33 @@ def export_images(
...
+        f += '<meta name="ocr-langs" content="unknown" />'
+        f += f'<meta name="ocr-number-of-pages" content="{len(self.pages)}" />'
+        f += '<meta name="ocr-capabilities" content="ocr_page ocr_carea ocr_par ocr_line ocrx_word" />'
+        f += "</head>"
+        f += "<body>"
+        for page_to_export in self.pages:
+            f += page_to_export.to_hocr(title=title)
+        f += "</body>"
+        f += "</html>"
+        return f
```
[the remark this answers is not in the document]
holtskinner: I get that you need all of the document information. But is there anything preventing accessing that information from an hocr converter file in the converters directory? It just seems strange to me to have all of this stuff in the wrapper classes when we have other convertors in separate locations.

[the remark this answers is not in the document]
galz10: Similar to the other export functions, they have functions inside the document wrapper. I just don't think having a file with one function is worth moving that logic out.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -267,13 +333,249 @@ def _table_rows_from_documentai_table_rows(
...
+
+    Args:
+        element (ElementWithLayout):
+            Required. an element with layout fields.
+        dimension (documentai.Document.Page.Dimension):
+            Required. Page dimension.
+        normalized (Boolean):
+            Required. Wether element.layout.bounding_poly is normalized
+        min (Boolean):
+            Required. Wether xy should be min
```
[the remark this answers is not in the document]
galz10: Same comment as below trying to reduce duplicate code

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
holtskinner: I guess it makes some sense. I'll take a look later to see if any of those functions could be refactored out to make the wrappers a little cleaner.

However, I would still prefer to see the hocr template in an xml somewhere instead of hard-coding it in the Python. If it's possible. https://github.com/googleapis/python-documentai-toolbox/pull/123#discussion_r1242316351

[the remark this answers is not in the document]
galz10: It doesn't look like there's an easy way to make a template version of this, since there are variables that need to change like title and self.pages so i don't know if there's an easy way to make it cleaner.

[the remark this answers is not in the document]
galz10: Done.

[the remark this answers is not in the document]
galz10: Fixed [#123 (comment) ](https://github.com/googleapis/python-documentai-toolbox/pull/123#discussion_r1242316351)

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -767,3 +769,27 @@ def export_images(
...
+            str:
+                A string hOCR version of the Document
+        """
+        environment = Environment(loader=FileSystemLoader("templates/"))
+        template = environment.get_template("hocr_xml_template.txt")
+        hocr_pages = ""
+        number_of_pages = len(self.pages)
+        for page_to_export in self.pages:
+            hocr_pages += page_to_export.to_hocr(title=title)
+
```
dizcology, line 790: This seems to say that the same title is shared across the whole document and every page?

In the hOCR specification must each page have a title?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -103,80 +113,90 @@ def sample_table_to_csv():
...
+    _text: Optional[str] = dataclasses.field(init=False, default=None)
+    _hocr_bounding_box: Optional[str] = dataclasses.field(init=False, default=None)
+
+    def __post_init__(self):
+        self.tokens = _get_children_of_element(
+            self.documentai_object, self._page.tokens
         )
 
+    @property
+    def text(self):
```
dizcology, line 177: Eventually let's upgrade `ElementWithLayout` to an abstract base class and move these duplicated codes into it.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -186,56 +206,101 @@ class FormField:
...
+        if self._text is None:
+            self._text = _text_from_layout(
+                layout=self.documentai_object.layout, text=self._page.text
+            )
+        return self._text
+
+    @property
+    def hocr_bounding_box(self):
+        if not self._hocr_bounding_box:
```
dizcology, line 298: Let's be consistent and go with `if self._hocr_bounding_box is None` throughout.

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
dizcology: To avoid calling the helper function too many times unnecessarily, in `_get_hocr_bounding_box` you can duplicate the return statement.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: will do

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +611,76 @@ class Page:
...
+            title (str):
+                Required. The title for ocr_page.
+
+        Returns:
+            str:
+                A string hOCR version of the documentai.Document.Page.
+        """
+        f = ""
+        pidx = self.documentai_object.page_number
+        f += f"<div class='ocr_page' lang='unknown' title='image \"{title}\";{self.hocr_bounding_box}'>"
```
dizcology, line 663: Why is there a hardcoded "image" substring in title here?

I also just noticed that we set `lang` to be unknown.  Is that a reasonable and accepted value for title too?  Perhaps we can either remove the `title` argument in `to_hocr` methods and just set them to be unknown?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +611,76 @@ class Page:
...
+        for bidx, block in enumerate(self.blocks):
+            f += f"<span class='ocr_carea' id='block_{pidx}_{bidx}' title='{block.hocr_bounding_box}'>"
+            for paridx, paragraph in enumerate(block.paragraphs):
+                f += f"<span class='ocr_par' id='par_{pidx}_{bidx}_{paridx}' title='{paragraph.hocr_bounding_box}'>"
+                for lidx, line in enumerate(paragraph.lines):
+                    line_text = html.escape(line.text)
+                    f += f"<span class='ocr_line' id='line_{pidx}_{bidx}_{paridx}_{lidx}' title='{line.hocr_bounding_box}'>{line_text}</span>"
+                    for tidx, token in enumerate(line.tokens):
+                        word_text = html.escape(token.text)
+                        f += f"<span class='ocrx_word' id='word_{pidx}_{bidx}_{paridx}_{lidx}_{tidx}' title='{token.hocr_bounding_box}'>{word_text}</span>"
```
dizcology, line 673: To be sure, is this structure of the `id` the hOCR convention?  Let's document this in one of the docstrings if not already done.

From the perspective of the XML, having these indices recorded in the `id` does not seem to provide very much value.  Must the `id` be meaning in anyway, or can we just use a short UUID?

holtskinner (approved):

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -310,27 +611,76 @@ class Page:
...
+        Returns:
+            str:
+                A string hOCR version of the documentai.Document.Page.
+        """
+        f = ""
+        pidx = self.documentai_object.page_number
+        f += f"<div class='ocr_page' lang='unknown' title='image \"{title}\";{self.hocr_bounding_box}'>"
+        for bidx, block in enumerate(self.blocks):
+            f += f"<span class='ocr_carea' id='block_{pidx}_{bidx}' title='{block.hocr_bounding_box}'>"
+            for paridx, paragraph in enumerate(block.paragraphs):
```
dizcology, line 666: Consider calling the children elements' `to_hocr` recursively, but if it not obvious how to easily do that, let's just track it in an internal clean up issue/bug.

google/cloud/documentai_toolbox/wrappers/page.py
dizcology: This still stands.  If i understand correctly, `dimension` is a property of `element_with_layout._page` and does not need to be an argument of this function.

google/cloud/documentai_toolbox/wrappers/page.py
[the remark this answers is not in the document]
dizcology: With the more recent changes, it seems to make sense to get rid of these functions?  For example, the `Page` class (and none of the other wrapper classes) could have cached properties `Page.blocks`, `Page.paragraphs` and so on that records these elements.

dizcology (approved):

google/cloud/documentai_toolbox/wrappers/page.py
galz10: removed title from page, i looked at the hocr documentation and the title of the document is not needed.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: I think the `id` has some meaning, both the hocr documentation and the client that will use this have the id in that format, i'll add some description about the id in the docstring.

google/cloud/documentai_toolbox/wrappers/page.py
galz10: Let's add it to internal clean up in buganizer.

PR #129, opened 2023-06-16
fix: Update noxfile.py

holtskinner:
- This is in line with how other client libraries/sample repos are set up


galz10 (approved):

dizcology (approved):
Looks like we are trying to keep this consistent with other library/sample repos.  My only question is whether such files could be automatically kept up to date.

PR #137, opened 2023-06-30
refactor: Refactor of hocr functions and fixing lost changes

holtskinner:
- Added Sample for hocr export
  - Fixed issue where Jinja Template wouldn't work in installed library.


holtskinner:
Note - the error for the Samples might actually need fixing in the document.py wrapper. Not sure if the paths all line up correctly for the template when using the compiled library

galz10 (approved):

dizcology (approved):

PR #139, opened 2023-07-06
refactor: Major refactoring of functions to improve readability, efficiency and follow standard practices.

holtskinner:
- Shouldn't impact any user-facing functions

BEGIN_COMMIT_OVERRIDE

fix: Internal refactoring to improve efficiency and readability. No external-facing changes.

END_COMMIT_OVERRIDE


google/cloud/documentai_toolbox/converters/config/bbox_conversion.py
```diff
@@ -130,27 +133,19 @@ def _convert_bbox_units(
...
-                final_coordinate = _normalize_coordinates(x, height)
-            else:
-                final_coordinate = _normalize_coordinates(x, width)
-
-    return final_coordinate * multiplier
+    pixel_conversion_rates = {
+        "pxl": 1,
+        "inch": 96,
+        "cm": 37.795,
+    }
```
dizcology, line 140: Can we put these in constants?

google/cloud/documentai_toolbox/converters/config/bbox_conversion.py
```diff
@@ -109,7 +108,11 @@ def _convert_to_pixels(x: float, conversion_rate: float) -> float:
 
 
 def _convert_bbox_units(
-    coordinate, input_bbox_units, width=None, height=None, multiplier=1
+    coordinate: float,
+    input_bbox_units: str,
```
dizcology, line 112: Can `input_bbox_units` be an enum instead of `str`?

google/cloud/documentai_toolbox/converters/config/block.py
```diff
@@ -0,0 +1,239 @@
...
+            schema_json.entity.normalized_vertices, "height", None
+        )
+        bounding_type = getattr(schema_json.entity.normalized_vertices, "type", None)
+        bounding_unit = getattr(schema_json.entity.normalized_vertices, "unit", None)
+        bounding_x = getattr(schema_json.entity.normalized_vertices, "x", None)
+        bounding_y = getattr(schema_json.entity.normalized_vertices, "y", None)
+
+        blocks: List[Block] = []
+        ens = _get_target_object(objects, entities)
+        for i in ens:
```
dizcology, line 183: The logic of this for loop and its nested if-else clauses is difficult to understand.  I get the impression that many of them could be simplified if a block `b`'s attributes have reasonable defaults.  For instance, can't

```
if bounding_width:
                b.bounding_width = _get_target_object(b.bounding_box, bounding_width)
```

be simplified to just
```
b.bounding_width = _get_target_object(b.bounding_box, bounding_width)
```
?

(Please ignore and file an internal clean up issue if this file was simply renamed.)

google/cloud/documentai_toolbox/converters/converter.py
```diff
@@ -15,9 +15,368 @@
...
+from google.cloud import documentai
+from google.cloud.documentai_toolbox import constants
+from google.cloud.documentai_toolbox.converters.config import bbox_conversion
+from google.cloud.documentai_toolbox.converters.config.block import Block
+from google.cloud.documentai_toolbox.utilities import gcs_utilities
+
+FILES_TO_IGNORE = [".DS_Store"]
+
+
+def _get_base_ocr(
```
dizcology, line 34: Are these simply moved from converter_helpers?  Let me know if these additions need a more detailed review.

dizcology (approved):

google/cloud/documentai_toolbox/converters/converter.py
holtskinner: These were refactored as well, and everything was moved into this file.

You can see just the refactoring work in this commit https://github.com/googleapis/python-documentai-toolbox/pull/139/commits/fb517fba8154e8e8da0553b8c8b3467a39061786

google/cloud/documentai_toolbox/converters/config/block.py
holtskinner: This was a refactoring that I did as well. So I'm good to make some changes as needed.

google/cloud/documentai_toolbox/converters/config/bbox_conversion.py
holtskinner: I agree that would make more sense. However, since this value is provided by the user in a JSON file, I wonder if there is a straightforward way to make that conversion? (Besides changing the config value to be an integer) Or maybe if it would be better to use constants?

google/cloud/documentai_toolbox/converters/config/bbox_conversion.py
holtskinner: Do you mean make the whole variable a constant? Or constants for the strings/Numbers?

PR #145, opened 2023-07-20
feat: Add export merged sharded Document proto

### LINKED ISSUE #156: Discrepancies in Properties and Format within hocr_document_template.xml.j2
I'm trying to create the hOCR file using a JSON file that I generated through DocumentAI. However, it seems that the hocr_document_template.xml.j2 template anticipates a distinct structure. To elaborate, it requires paragraphs followed by nested lines within those paragraphs, with each line further containing nested words. In contrast, the JSON produced by DocumentAI doesn't adhere to this format. Moreover, the hocr_document_template.xml.j2 template utilizes dissimilar property names like hocr_bounding_box, which are absent in the JSON file (attached).

Adding to this, the code repository displays a templates folder at google/cloud/documentai_toolbox/templates/. Nonetheless, upon installing google.cloud.documentai_toolbox via pip, the templates folder appears to be missing.

[wikipedia-sample.json.txt](https://github.com/googleapis/python-documentai-toolbox/files/12292698/wikipedia-sample.json.txt)

holtskinner:
- `to_documentai_document` exports a documentai Document proto from all of the shards in the wrapped Document
- Added fix for template dependencies 
- Fixes #156


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -771,3 +802,26 @@ def export_hocr_str(self, title: str) -> str:
         template = environment.get_template("hocr_document_template.xml.j2")
         content = template.render(pages=self.pages, title=title)
         return content
+
+    def to_documentai_document(self) -> documentai.Document:
```
dizcology, line 806: Maybe rename the method so the *merging* aspect is more obvious.

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -771,3 +802,26 @@ def export_hocr_str(self, title: str) -> str:
...
+        Returns:
+            documentai.Document:
+                Document with all shards merged and text offsets applied.
+        """
+        if len(self.shards) == 1:
+            return self.shards[0]
+
+        merged_document = documentai.Document(text=self.text, pages=[], entities=[])
+        for shard in self.shards:
+            _apply_text_offset(
```
dizcology, line 820: Since the current implementation of `_apply_test_offset` modifies the shard in-place, after calling `to_documentai_document` the `Document` objects is mutated too.  This side effect is undesirable since we cannot be sure whether the same `Document` object needs to be used for other purposes after the call.

dizcology (changes requested):

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -284,6 +284,37 @@ def _dict_to_bigquery(
     )
 
 
+def _apply_text_offset(documentai_object: object, text_offset: int) -> None:
```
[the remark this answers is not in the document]
holtskinner: I like the idea of using lenses, haven't heard of it before.

The original implementation was done by ffieldhaus@

This code treats each object like arbitrarily deeply nested dictionaries, it looks for every nested instance of `text_segments` and adds the offset to the start and end values.

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: Ah, that's correct. I didn't intend to modify everything in place, but my original implementation didn't seem to modify as expected with all of the recursion involved. I'll try making a full copy

google/cloud/documentai_toolbox/wrappers/document.py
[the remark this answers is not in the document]
dizcology: Looking at this more closely, I am unsure if lenses can help us simplify the code here.  I think we should still make the annotation a little bit more specific, though.

dizcology (approved):

PR #171, opened 2023-09-25
feat: Update Toolbox for OCR 2.0 features

holtskinner:
- Refactor Page Element classes to reduce duplication
- Add Math Elements & Symbol (Character) elements
- Add more examples of layout elements to quickstart code sample
- Context: https://cloud.google.com/document-ai/docs/enterprise-document-ocr


google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -431,15 +373,19 @@ class Page:
     text: str = dataclasses.field(init=False, repr=False)
     page_number: int = dataclasses.field(init=False, repr=False)
     form_fields: List[FormField] = dataclasses.field(init=False, repr=False)
+    symbols: List[Symbol] = dataclasses.field(init=False, repr=False)
+    tokens: List[Token] = dataclasses.field(init=False, repr=False)
```
holtskinner, line 377: This should have been added previously

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -111,25 +104,20 @@ def __post_init__(self, document_text) -> None:
...
-    .
-
-    Attributes:
-        documentai_object (google.cloud.documentai.Document.Page.Token):
-            Required. The original google.cloud.documentai.Document.Page.Token object.
-        _page (Page):
-                    Required. The Page object.
-    """
+class PageElement:
+    """Base class for representing a wrapped Document AI Page element (Symbol, Token, Line, Paragraph, Block)."""
```
dizcology, line 108: If this class is meant to be used only as a base class and not instantiated by the library, maybe make it an abstract class.  (Maybe also consider renaming it to something like "BasePageElement" or "PageElementBase" to ensure that future developers of the library can easily recognize its role.)

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -147,133 +138,75 @@ def hocr_bounding_box(self):
...
 
-    Attributes:
-        documentai_object (google.cloud.documentai.Document.Page.Line):
-            Required. The original google.cloud.documentai.Document.Page.Line object.
-        _page (Page):
-            Required. The Page object.
-    """
+    @property
+    def hocr_bounding_box(self):
+        return None
```
dizcology, line 147: Does this mean that symbols are not represented in the hocr format?  If so please add some code comments here to document that.

Same comment about math formula below.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -16,21 +16,14 @@
...
-from typing import cast, List, Optional, Union
+from typing import cast, List, Optional
 
 import pandas as pd
 
 from google.cloud import documentai
 from google.cloud.documentai_toolbox.constants import ElementWithLayout
 from google.cloud.documentai_toolbox.utilities import docai_utilities
 
-ChildrenElements = Union[
```
dizcology, line 27: Just a comment, no change required:

Below some of the occurrences of `ChildrenElements` are replaced with `List[ElementWithLayout]`.  This should be fine, but I want to call out the differences here:

A list of mixed type elements such as `[a_paragraph, a_block]` would be accepted as a `List[ElementWithLayout]`, but not accepted as a `ChildrenElements`.  So these two are not complete interchangeable.  (Such a list should never occur naturally.)

dizcology (approved):

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: That's entirely true, I just didn't think there was a clean way to notate this. Considering we need to use some kind of super class `ElementWithLayout` as the base `documentai_object`

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: Thanks for the suggestion!

PR #169, opened 2023-09-23
fix: Change `ocr_line` `<span>` to include all `ocr_word`

ralscha:
Fixes the xml for ocr_line. The span of ocr_line should enclose all spans of ocr_word

PR #161, opened 2023-09-01
fix: Add handling for documents missing all layout elements.

holtskinner:
- Bounding boxes for empty elements will show up in output as `0,0,0,0`

Fixes #160 🦕

PR #176, opened 2023-10-03
fix: `docai_utilities.py` to return `Optional`

holtskinner:
- Should resolve customer reported issue in support case #47169701 relating to duplicate/inaccurate elements in hOCR output
- Followup to:
  - #161
  - #169


google/cloud/documentai_toolbox/utilities/docai_utilities.py
```diff
@@ -37,8 +37,8 @@ def get_bounding_box(
             Bounding box coordinates in order (top, left, bottom, right).
             Returns `0, 0, 0, 0` if `bounding_poly.normalized_vertices` is empty.
     """
-    if not bounding_poly.normalized_vertices:
-        return 0, 0, 0, 0
+    if not bounding_poly or not bounding_poly.normalized_vertices:
```
dizcology, line 40: Update the docstring for this changed behavior.

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -95,9 +95,11 @@ def crop_image(
         if not documentai_page.image:
             raise ValueError("Document does not contain images.")
 
-        top, left, bottom, right = docai_utilities.get_bounding_box(
+        bbox = docai_utilities.get_bounding_box(
             bounding_poly=self.documentai_object.page_anchor.page_refs[0].bounding_poly,
             page_dimension=documentai_page.dimension,
         )
+        if not bbox:
```
dizcology, line 102: In this case let's use `if bbox is None`.  Also perhaps move this check to immediately after we get `bbox`.

tests/unit/test_document.py
```diff
@@ -648,6 +656,29 @@ def test_export_images(get_bytes_images_mock):
...
 
 
+def test_export_images_empty_bounding_box(get_bytes_images_mock):
+    output_path = "resources/output/"
+
+    doc = document.Document.from_gcs(
+        gcs_bucket_name="test-directory", gcs_prefix="documentai/output/123456789/0"
+    )
+
+    del (
```
dizcology, line 666: Perhaps put these `doc = ...; del(...` lines of code into a factory function, say, `def make_document_without_bbox`, and each test that needs a document of that shape can call the function to get a copy of it.

dizcology (changes requested):

holtskinner:
> If possible, please also give some explanations about the duplicate/inaccurate elements in hOCR output in the PR description.

I will once I get that information

google/cloud/documentai_toolbox/utilities/docai_utilities.py
holtskinner: Missed this.

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: Done, although I'm not sure of your second comment. This is already right after we get bbox.

holtskinner:
Note: In the customers code, they use this whenever the document is blank. Not sure if this is a standard structure for blank hOCR documents, but could be good to look into.

```py
    def _build_empty_hocr(self):
        return '''<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="unknown" lang="unknown">
<head>
<title>hocr</title>
<meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
<meta name="ocr-system" content="Document AI OCR" />
<meta name="ocr-langs" content="unknown" />
<meta name="ocr-number-of-pages" content="1" />
<meta name="ocr-capabilities" content="ocr_page ocr_carea ocr_par ocr_line ocrx_word" />
</head>
<body>
<div class='ocr_page' lang='unknown' title='bbox 0 0 0 0'>
<span class='ocr_carea' id='block_1_0' title='bbox 0 0 0 0'>
<span class='ocr_par' id='par_1_0_0' title='bbox 0 0 0 0'>
<span class='ocr_line' id='line_1_0_0_0' title='bbox 0 0 0 0'></span>
<span class='ocrx_word' id='word_1_0_0_0_0' title='bbox 0 0 0 0'></span>
</span>
</span>
</div>
</body>
</html>'''
```

dizcology (approved):

PR #194, opened 2023-11-07
fix: Add Backwards Compatibility for incorrect hOCR Format

### LINKED ISSUE #193: Support different formats in method export_to_hocr
Hi,

I have found a backward incompatibility regarding to the change included in the 2a0 version. When using the wrapper.document.py. export_hocr_str() method, the hocr output format is different to previous versions.

This fix changes the embedding of the "ocr_line" class in the hocr output, including now the "ocrx_word" classes that compounds the line.  This fix makes the output more logical and according to the spec, however all trained processors based in the previous hocr format are now misfunctioning.

I want to propose a solution where the method allows to export the hocr in the previous format, keeping the current format as the default. Something the way:

```
def export_hocr_str(self, title: str, inline_words = true : bool) -> str:
        environment = Environment(
            loader=PackageLoader("google.cloud.documentai_toolbox", "templates")
        )
        template = environment.get_template("hocr_document_template_inline_words.xml.j2")
        if not inline_words:
           template = environment.get_template("hocr_document_template.xml.j2")
        content = template.render(pages=self.pages, title=title)
        return content
```

Not the best code, as i consider that maybe the j2 allows this kind of parameterization to change its output, but the most straightforward than i can think of.

Thanks for your attention.

holtskinner:
- Fixes #193 - Adds `inline_words` parameter to `export_hocr_str()` to support previous (incorrect) format.


dizcology:
Let's hold this, pending some internal discussions.  The only situation when we support and maintain multiple templates should be if they all conform to the hOCR spec, e.g. different spec version, or the spec allows some variations and there are multiple common and popular options by the users.

holtskinner:
Closing out, should no longer be needed

PR #195, opened 2023-11-07
fix: Updates to hOCR Template to follow hOCR Spec

holtskinner:
- Added validation in testing with https://github.com/kba/hocr-spec-python


google/cloud/documentai_toolbox/templates/hocr_document_template.xml.j2
```diff
@@ -6,8 +6,9 @@
 <meta http-equiv="Content-Type" content="text/html;charset=utf-8" />
 <meta name="ocr-system" content="Document AI OCR" />
 <meta name="ocr-langs" content="unknown" />
+<meta name="ocr-scripts" content="unknown" />
```
dizcology, line 9: If `.j2` template files support code comments, please include some links to the source of the format used here, including the hOCR spec version/date if possible.

dizcology (approved):

PR #200, opened 2023-11-14
fix: Prevent sorting entities labeled in Document AI Workbench

### LINKED ISSUE #199: Sorting function doesn't need int() conversion
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search the issues already opened: https://github.com/googleapis/python-documentai-toolbox/issues
  - Search StackOverflow: https://stackoverflow.com/questions/tagged/google-cloud-platform+python

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS type and version:
  - Python version: `python --version`
  - pip version: `pip --version`
  - `google-cloud-documentai-toolbox` version: `pip show google-cloud-documentai-toolbox`

#### Steps to reproduce

  1. ?
  2. ?

#### Code example

```python
wrapped_document = document.Document.from_gcs(
    gcs_bucket_name=gcs_bucket_name, gcs_prefix=gcs_prefix
)
```

#### Stack trace
```
File ~/miniconda/lib/python3.10/site-packages/google/cloud/documentai_toolbox/wrappers/document.py:70, in _entities_from_shards.<locals>.<lambda>(x)
     67     page_offset += len(shard.pages)
     69 if len(result) > 1 and result[0].documentai_object.id:
---> 70     result.sort(key=lambda x: int(x.documentai_object.id))
     71 return result

ValueError: invalid literal for int() with base 10: 'e6a445871928ed89'
```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

lababidi:
remove int because non base 10 objects are used as id by default. I was getting an error because the id in my google bucket of objects was not a numeric but hexadecimal. 

Fixes #199 🦕


lababidi:
@parthea Any Guidance on getting this through the process would be appreciated :)

holtskinner (approved):
Thanks for the catch! I didn't realize these values could be Hexadecimal.

holtskinner:
Refactored how loading and sorting of entities is done to handle this edge case.

Sorting will only occur if all entities have a numeric id, and they will be sorted as integers.

dizcology (approved):
Please add some code comments documenting the rationale behind the conditional sorting.

PR #205, opened 2023-11-28
fix: Implement lazy-loading of properties to speed up initialization

### LINKED ISSUE #198: Reduce initialization time for a wrapped `Document`
When a wrapped `Document` is created, all page elements such as `Block`, `Paragraph`, `Line`, `Token` have their text initialized at creation time, which takes a significant amount of time for large documents (and increases the overall object size).

It would be very helpful for users if this time could be reduced while still maintaining the ease of accessing text directly from the page elements and not taking significant time to access this field.

---

Possible way to improve this is to use the `@property` decorator for the "child" element fields like [`Block.paragraphs`](https://github.com/googleapis/python-documentai-toolbox/blob/0baf3efaae1d845077b82d47bc22f0e6a537e22a/google/cloud/documentai_toolbox/wrappers/page.py#L239C5-L239C5)

And the "child" elements of `Page` such as [`Page.form_fields`](https://github.com/googleapis/python-documentai-toolbox/blob/0baf3efaae1d845077b82d47bc22f0e6a537e22a/google/cloud/documentai_toolbox/wrappers/page.py#L431)

Will require some testing to find where the latency is highest.

holtskinner:
Fixes #198 🦕


holtskinner:
In testing, the initialization time is about 1/3 of the previous time for large documents.

`43.9 s → 15.1 s`

## Old Version
Screenshot 2023-11-28 at 12 49 18 PM

## New Version

Screenshot 2023-11-28 at 12 50 34 PM

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -33,28 +33,40 @@ class Table:
...
-    def __post_init__(self, document_text) -> None:
-        self.header_rows = _table_rows_from_documentai_table_rows(
-            table_rows=list(self.documentai_object.header_rows), text=document_text
-        )
-        self.body_rows = _table_rows_from_documentai_table_rows(
-            table_rows=list(self.documentai_object.body_rows), text=document_text
-        )
+    @property
+    def body_rows(self):
+        if self._body_rows is None:
```
dizcology, line 55: Looks like we are adding the same pattern in many places.  You might consider using this (which I think replaces the need for fields like `self._body_rows` here): https://docs.python.org/3/library/functools.html#functools.cached_property

dizcology (approved):

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: I agree, I thought that would be a better/cleaner option, but `@cached_property` allows for writes, which we don't want to allow on these properties.

From what I've found, this is the recommended procedure for implementing it.

PR #223, opened 2023-12-12
feat: Added `Document.from_gcs_multi()` method to get multiple wrapped documents from a GCS directory.

### LINKED ISSUE #214: Allow Importing Multiple Documents from a single GCS Path
Inspired by https://stackoverflow.com/a/77609221/6216983

The[`from_gcs()`][2] method can only create a single Wrapped Document from a single document output in GCS.

It could be simpler for users if this method could output a list of Wrapped Documents from a GCS path similar to [`from_batch_process_operation()`][3] and [`from_batch_process_metadata()`][4].

However, this would be backwards incompatible, so an alternative is to create a new method for this functionality. (Not sure what the best name would be)

[2]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_gcs

[3]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_batch_process_operation

[4]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_batch_process_metadata

### LINKED ISSUE #15828: Allow Importing Multiple Documents from a single GCS Path
Inspired by https://stackoverflow.com/a/77609221/6216983

The[`from_gcs()`][2] method can only create a single Wrapped Document from a single document output in GCS.

It could be simpler for users if this method could output a list of Wrapped Documents from a GCS path similar to [`from_batch_process_operation()`][3] and [`from_batch_process_metadata()`][4].

However, this would be backwards incompatible, so an alternative is to create a new method for this functionality. (Not sure what the best name would be)

[2]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_gcs

[3]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_batch_process_operation

[4]: https://cloud.google.com/python/docs/reference/documentai-toolbox/latest/google.cloud.documentai_toolbox.wrappers.document.Document#google_cloud_documentai_toolbox_wrappers_document_Document_from_batch_process_metadata

holtskinner:
Fixes googleapis/google-cloud-python#15828 🦕

- Note: `from_gcs()` takes in a GCS directory, but it only works for a single sharded document from a single input document source.
- In a GA release, it would be a better practice to have `from_gcs()` take in any GCS directory and output a list of Wrapped Documents. But this would be a backwards-incompatible change now.
  - Not sure if it's possible/advisable to have two possible return types for `from_gcs()` and just have it return a list when there are multiple Wrapped Documents?


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -481,6 +481,35 @@ def from_gcs(
...
+
+        Args:
+            gcs_bucket_name (str):
+                Required. The gcs bucket.
+
+                Format: Given `gs://{bucket_name}/{optional_folder}/{operation_id}/` where `gcs_bucket_name={bucket_name}`.
+            gcs_prefix (str):
+                Required. The prefix to the location of the target folder.
+
+                Format: Given `gs://{bucket_name}/{optional_folder}/{target_folder}` where `gcs_prefix={optional_folder}/{target_folder}`.
```
dizcology, line 500: Can we use the same "format" example in the doc string for `gcs_bucket` and `gcs_prefix`?

google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -481,6 +481,35 @@ def from_gcs(
...
+        Returns:
+            List[Document]:
+                A List of documents from gcs.
+        """
+        return [
```
dizcology, line 505: This is assuming that every sub-directory in the parent (bucket+prefix) is acceptable by `Document.from_gcs`, which I think has some validation checks.

What is the desired behavior here if some of those sub-directories causes `Document.from_gcs` to fail?  Should the whole `list_from_gcs` call fail, or should it return just whatever is successfully loaded (maybe along with some metadata explaining the failed items)?

(Let's include unit tests for the case where one of the sub-directories contains wrong files.)

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: Done

google/cloud/documentai_toolbox/wrappers/document.py
holtskinner: `list_document_tree()` only outputs directories that contain JSON files. (However, it doesn't directly validate if they are `Document` JSON format.)

`.from_gcs()` will throw a `ValueError` if a file is invalid. Or the `documentai.Document.from_json()` method will throw an exception.

It would probably make sense for this batch case to return whatever was successfully loaded and just skip files that throw exceptions. (And printing out or returning the failed files.)

holtskinner:
Closing out this PR. I don't think it makes sense to add this method because only the end user will know how exactly the GCS directories are set up. For example, if the documents in each directory are shards of the same document or should be treated as separate documents.

Unless the documents are being imported from the Batch Process output, in that case, the user can use `from_batch_process_operation()` or `from_batch_process_metadata()`.

PR #261, opened 2024-02-23
feat: Added `Document.from_gcs_uri()` method to allow importing of a single Document JSON

holtskinner:


google/cloud/documentai_toolbox/wrappers/document.py
```diff
@@ -482,7 +487,27 @@ def from_gcs(
             Document:
                 A document from gcs.
         """
-        shards = _get_shards(gcs_bucket_name=gcs_bucket_name, gcs_prefix=gcs_prefix)
+
+        if bool(gcs_bucket_name) == bool(gcs_uri) or bool(gcs_prefix) == bool(gcs_uri):
```
dizcology, line 491: This kind of checks become difficult to maintain and debug.  Technically if both `gcs_bucket_name` and `gcs_uri` are omitted by the caller, the error will not be raised.

To avoid such difficulties and confusion, I would recommend simply introducing a new method, maybe `Document.from_gcs_uri`.

holtskinner: I agree the checks are a bit confusing to maintain. I mostly wanted to keep the same method to avoid user confusion about having multiple methods. But I can change this around.

dizcology (approved):

PR #316, opened 2024-06-13
fix: Refactor page.py to improve performance and organization

### LINKED ISSUE #312: High latency exporting to hocr
Hello,

Coming from a Google Support Case 51622001: High latency exporting to hocr, which has derived to this issue.

#### Details
In the method from google.cloud.documentai_toolbox import document as documentai_document_wrapper

```python
documentai_document_wrapper.Document.from_documentai_document(
        documentai_document=result.document
    ).export_hocr_str(title="title")
```

When transforming tables, the latency takes from 30 to 50 seconds, depending on the complexity of the page (high data in table formats).

Looking for any type of optimization.

#### Environment details

  - OS type and version: GCP cloudshell 
  - `google-cloud-documentai-toolbox` version: 0.13.3a0

#### Steps to reproduce

  1. create venv with the provided requirements.txt
  2. execute python3 main-hocr.py test.pdf

#### Code example

```python
 request = documentai.ProcessRequest(
      name=resource_name,
      raw_document=raw_document,
      process_options=process_options,
  )

  start = time.time()
  result = client.process_document(request=request)
  print(f"process_document {(time.time() - start)}")

  start = time.time()
  wrapped_document = documentai_document_wrapper.Document.from_documentai_document(
      documentai_document=result.document
  )
  print(f"wrapped_document {(time.time() - start)}")

  start = time.time()
  hocr_result = wrapped_document.export_hocr_str(title="hocr")
  print(f"export_hocr_str {(time.time() - start)}")
```

#### Stack trace
N/A, the execution is correct, but the latency takes 35 seconds long

Attached sources to reply the test:
[sources.zip](https://github.com/user-attachments/files/15807105/sources.zip)
* main-hocr.py, with the full code of the example
* requirements.txt
* test.pdf, file to process with documentai: ocr plus hocr

Thanks!

holtskinner:
Fixes #312 🦕

Improves upon the hOCR processing improvements made in #313

```
         22788704 function calls (19761621 primitive calls) in 6.297 seconds
```

### DIFF
--- a/google/cloud/documentai_toolbox/wrappers/page.py
+++ b/google/cloud/documentai_toolbox/wrappers/page.py
@@ -188,16 +188,22 @@ class _BasePageElement(ABC):
             List[_BasePageElement]:
                 A list of wrapped children that are inside an element.
         """
-        return [
-            child
-            for child in potential_children
-            if self._text_segment.start_index
-            <= child._text_segment.start_index
-            < self._text_segment.end_index
-            and self._text_segment.start_index
-            < child._text_segment.end_index
-            <= self._text_segment.end_index
-        ]
+        start_index = self._text_segment.start_index
+        end_index = self._text_segment.end_index
+
+        children = []
+        for child in potential_children:
+            child_start_index = child._text_segment.start_index
+            child_end_index = child._text_segment.end_index
+
+            if child_start_index >= end_index:
+                break
+            if (
+                start_index <= child_start_index < end_index
+                and start_index < child_end_index <= end_index
+            ):
+                children.append(child)
+        return children
 
 
 @dataclasses.dataclass
@@ -344,7 +350,7 @@ def _text_from_layout(layout: documentai.Document.Page.Layout, text: str) -> str
 
     Args:
         layout (documentai.Document.Page.Layout):
-            Required. an element with layout fields.
+            Required. An element with layout fields.
         text (str):
             Required. UTF-8 encoded text in reading order
             of the `documentai.Document` containing the layout element.
@@ -353,6 +359,8 @@ def _text_from_layout(layout: documentai.Document.Page.Layout, text: str) -> str
         str:
             Text from a single element.
     """
+    if not layout.text_anchor or not layout.text_anchor.text_segments:
+        return ""
 
     # Note: `layout.text_anchor.text_segments` are indexes into the full Document text.
     # https://cloud.google.com/document-ai/docs/reference/rest/v1/Document#textsegment
--- a/google/cloud/documentai_toolbox/wrappers/page.py
+++ b/google/cloud/documentai_toolbox/wrappers/page.py
@@ -196,21 +196,6 @@ class _BasePageElement(ABC):
             List[_BasePageElement]:
                 A new list containing only the wrapped page elements that are fully
                 contained within this element, maintaining their original order.
-
-        Raises:
-            TypeError: If `potential_children` is not a list or contains elements that are not of type `_BasePageElement`.
-
-        Example:
-        ```
-        page_element = PageElement(text_segment=TextSegment(0, 100))
-        potential_children = [
-            PageElement(text_segment=TextSegment(10, 20)),  # Inside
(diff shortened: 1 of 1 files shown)


google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -44,37 +44,54 @@ class Table:
...
+                Required. A documentai.Document.Page.Table.TableRow.
+
+        Returns:
+            List[List[str]]:
+                A list of table rows.
+        """
+        return [
+            [
+                _text_from_layout(cell.layout, self._page._document_text).replace(
+                    "\n", ""
```
dizcology, line 88: I know this already existed before this refactoring, but is removing all new lines (even if in the middle of the cell's content) what the user would want?  I imagine it is possible to have multi-line text in the headers.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -95,22 +112,37 @@ class FormField:
...
+    Args:
+        text (str):
+            Required. UTF-8 encoded text in reading order
+            from the document.
+
+    Returns:
+        str:
+            Text without trailing spaces/newlines
+    """
+    return text.strip().replace("\n", " ")
```
dizcology, line 143: Let's document this behavior in the docstring, it is not only removing space characters.

On the other hand, how can we be sure that the user does not car about leading/trailing white spaces (removed by `strip`) or newline characters in the middle of the text.

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -119,24 +151,60 @@ class _BasePageElement(ABC):
...
         return _get_hocr_bounding_box(
-            element_with_layout=self.documentai_object,
-            page_dimension=self._page.documentai_object.dimension,
+            self.documentai_object, self._page.documentai_object.dimension
         )
 
+    @cached_property
+    def _text_segment(self) -> documentai.Document.TextAnchor.TextSegment:
+        """
+        Page element text segment.
```
dizcology, line 174: Update this docstring to be more informative.  

The implementation is just a one-liner getting the first text segment, unless this is used in many places why not just inline these attribute accesses?

google/cloud/documentai_toolbox/wrappers/page.py
```diff
@@ -119,24 +151,60 @@ class _BasePageElement(ABC):
...
+        """
+        return self.documentai_object.layout.text_anchor.text_segments[0]
+
+    def _get_children_of_element(
+        self, potential_children: List["_BasePageElement"]
+    ) -> List["_BasePageElement"]:
+        """Returns a list of children inside element.
```
dizcology, line 181: Document this with more information about how we use the start/end indices to filter out some children.

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: I agree, this probably shouldn't remove the newlines by default. I think this was just for improving formatting with output like CSV, because it wouldn't always turn out properly.

Which would be preferred, removing this `.replace()` call and making a possibly breaking change, or leaving as is?

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: It's true, we don't know for sure that the user needs these removed. It was added as a convenience because many form field names/values are misshapen when output from DocAI.

Should this be removed?

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: This is specifically brought out into a method so it can be a cached property. This field can be accessed multiple times, and this change resulted in a major speed improvement when exporting hocr.

google/cloud/documentai_toolbox/wrappers/page.py
holtskinner: Done - Gemini to the rescue

dizcology (approved):

PR #333, opened 2024-07-11
fix: Add support for Classifier entities

### LINKED ISSUE #332: `Document.entities` field is unusable when using data from Classifier output
Hello,

The wrapped_document, when using `document.from_batch_process_metadata` (or any other methods) will be missing entities field when using data from the Classifier. 

When using output of splitter, everything works fine.
But with classifier - you wont get any important information like type and confidence.

```
from google.cloud.documentai_toolbox import document
import os

doc = document.Document.from_document_path(os.path.join(os.path.dirname(__file__), "output-document_split.json"))
print(doc.entities)
doc = document.Document.from_document_path(os.path.join(os.path.dirname(__file__), "output-document_classify.json"))
print(doc.entities)
```

[output-document_split.json](https://github.com/user-attachments/files/16169761/output-document_split.json)
[output-document_classify.json](https://github.com/user-attachments/files/16169762/output-document_classify.json)

holtskinner:
- Added support for Custom Classifier entities
- Added Unit Tests for Classifier output
- Added input validation for `Document.split_pdf()`

Fixes #332 🦕


google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -57,28 +57,33 @@ class Entity:
...
+    # Only Populated for Splitter Output
+    end_page: Optional[int] = dataclasses.field(init=False, default=None)
 
     _image: Optional[Image.Image] = dataclasses.field(init=False, default=None)
 
     def __post_init__(self, page_offset: int) -> None:
         self.type_ = self.documentai_object.type_
-        self.mention_text = self.documentai_object.mention_text
+
+        if self.documentai_object.mention_text:
```
dizcology, line 73: Is the point here that if mention_text is the empty string, we want to keep self.mention_text as `None`, so that we will be able to detect this by checking `self.mention_text is None` downstream?

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -57,28 +57,33 @@ class Entity:
...
     type_: str = dataclasses.field(init=False)
-    mention_text: str = dataclasses.field(init=False, default="")
-    normalized_text: str = dataclasses.field(init=False, default="")
+    mention_text: Optional[str] = dataclasses.field(init=False, default=None)
+    normalized_text: Optional[str] = dataclasses.field(init=False, default=None)
 
-    start_page: int = dataclasses.field(init=False)
-    # Only Populated for Splitter/Classifier Output
-    end_page: int = dataclasses.field(init=False)
+    # Not populated in Classifier output
```
dizcology, line 63: Add this to the docstring instead.

Also the docstring needs to be updated for all the changes here such as field type.

google/cloud/documentai_toolbox/wrappers/entity.py
```diff
@@ -57,28 +57,33 @@ class Entity:
...
+    start_page: Optional[int] = dataclasses.field(init=False, default=None)
+    # Only Populated for Splitter Output
```
dizcology, line 65: Some of these conditions become quite subtle.  Should each Document remember if it came from a Splitter or Classifier or...?  We could add some metadata for it if it helps validation.

tests/unit/test_entity.py
```diff
@@ -68,6 +68,18 @@ def test_Entity_splitter():
...
+    documentai_entity = documentai.Document.Entity(
+        type_="clinical_notes",
+        id="0",
+        confidence=0.99878639,
+    )
+    wrapper_entity = entity.Entity(documentai_entity)
+    assert wrapper_entity.type_ == "clinical_notes"
+    assert wrapper_entity.documentai_object.id == "0"
+    assert round(wrapper_entity.documentai_object.confidence, 8) == 0.99878639
+
```
dizcology, line 81: Maybe test some of the fields as well, for example if a field must remain `None` if coming from Classifier.

dizcology (approved):

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: Yes, that's why. Depending on how the output JSON is formatted, the `mention_text` field can be either `None` or the empty string and we want to keep it as `None` for an empty input.

google/cloud/documentai_toolbox/wrappers/entity.py
holtskinner: They do become subtle. I don't think the Document needs to know explicitly what type of processor it came from. This isn't output anywhere in the Document.JSON

dizcology (approved):

PR #390, opened 2025-11-18
feat: Add support for Python 3.14

daniel-sanche:
This PR adds Python 3.14 as a supported version, and adds new CI tests against the new version

To support 3.13 and 3.14, I had to increase some of the dependency bounds in the setup.py, but the tests still pass with the new versions

pre-release deps test will fail until the release of [numba==0.63](https://pypi.org/project/numba/#history), with 3.14 support


holtskinner (dismissed):

mrfaizal (approved):

daniel-sanche:
This is currently blocked until the `numba` dependency supports 3.14. It looks like they put out a [release candidate build](https://pypi.org/project/numba/#history) Nov 20

